PR #205, opened 2023-03-20
Add parameter for merge order and integration order

### LINKED ISSUE #201: Add parameter for specifying merge order and test with integrated ScPCA data
Originally posted in https://github.com/AlexsLemonade/scpca-nf/issues/275, we want to be able to incorporate the addition of `auto.merge` or randomizing the order of libraries prior to integration. However, we've never actually tested this parameter or used it. We should add that as a parameter to the workflow here and then re-run using the ScPCA datasets as the input so that we can generate reports with changes in integration order and evaluate if the integration order changes the results.

allyhawkins:
Closes #201 

This PR adds two new parameters to the integration workflow, one for specifying the merge order and one for specifying the integration order specifically for use with `fastmnn`. 

The first parameter, `--random_merge` that I added is a flag that can be used that will result in randomly shuffling the order of the SCE objects in the SCE list prior to creating the merged object. The merged object will be returned in the order that the SCE list was provided so using this flag results in a randomly ordered merged object. This object is then used as input into the integration scrips. Then for a tool like `fastmnn`, where the input order determines the integration order, unless otherwise specified, the randomized input that was used during merging will also be used for integration. Harmony does not have the option to specify the integration order and appears to use the order provided in the input object, so to ensure that the objects are all in the same order, I applied the randomization during the merge step rather than at the integration step. 

The second parameter, `--fastmnn_auto_merge` can be used only for integrating with `fastmnn` to test the use of `auto.merge=TRUE`. This means that objects will be integrated in an order that maximizes the number of mutual nearest neighbors between each pair rather than in the input order. This option is not applicable to other methods which is why I made it a `fastmnn` specific flag. Doing this means you can do both `--random_merge` and `--fastmnn_auto_merge` to compare random harmony vs. auto fastmnn or you can do `--ramdom_merge` only to compare the random input in both. To me this was the best way to accommodate the different ways both of these methods deal with integration order, but if others have other ideas I'm definitely open to that. 

In doing this, I also added notes to the metadata of the merged or integrated SCE objects to note the merge or integration order. Then in the integration report, I added messages that printed out what type of merging and integration was being used for that report. I attached an example of one of the reports for reference. 

For the snakefile, I added the new flags to the config file, but don't set them by default. When running the snakefile to test the new additions, I am running with the flags set and changing the results directory so as to separate results from when we don't enable random merge or auto integration. 
```
snakemake -c4 --use-conda --configfile config/config-scpca.yaml --config random_merge="--random_merge" fastmnn_auto_merge="--fastmnn_auto_merge" results_dir="results/scpca-random-auto" 
```

[SCPCP000007_integration_report.html.zip](https://github.com/AlexsLemonade/sc-data-integration/files/11020118/SCPCP000007_integration_report.html.zip)

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -46,6 +46,7 @@ rule merge_sces:
           --merged_sce_dir "{output}" \
           --num_hvg {config[num_hvg]} \
           --subset_hvg \
+          --seed {config[seed]} \
           {config[random_merge]} \
           &> {log}
         """
--- a/Snakefile
+++ b/Snakefile
@@ -41,7 +41,7 @@ rule merge_sces:
           --library_file "{input.processed_tsv}" \
           --add_celltype {config[add_celltype]} \
           --celltype_info "{config[celltype_file]}" \
-          --random_merge {config[random_merge]} \
+          {config[random_merge]} \
           --grouping_var {config[grouping_var]} \
           --groups_to_integrate "{config[groups_to_integrate]}" \
           --merged_sce_dir "{output}" \
--- a/Snakefile
+++ b/Snakefile
@@ -41,12 +41,12 @@ rule merge_sces:
           --library_file "{input.processed_tsv}" \
           --add_celltype {config[add_celltype]} \
           --celltype_info "{config[celltype_file]}" \
-          {config[random_merge]} \
           --grouping_var {config[grouping_var]} \
           --groups_to_integrate "{config[groups_to_integrate]}" \
           --merged_sce_dir "{output}" \
           --num_hvg {config[num_hvg]} \
           --subset_hvg \
+          {config[random_merge]} \
           &> {log}
         """
 
--- a/Snakefile
+++ b/Snakefile
@@ -90,6 +90,7 @@ rule integrate_fastmnn:
           --method fastMNN \
           --seed {config[seed]} \
           --corrected_only \
+          {config[fastmnn_auto_merge]} \
           &> {log}
         """
 
--- a/Snakefile
+++ b/Snakefile
@@ -41,6 +41,7 @@ rule merge_sces:
           --library_file "{input.processed_tsv}" \
           --add_celltype {config[add_celltype]} \
           --celltype_info "{config[celltype_file]}" \
+          --random_merge {config[random_merge]} \
           --grouping_var {config[grouping_var]} \
           --groups_to_integrate "{config[groups_to_integrate]}" \
           --merged_sce_dir "{output}" \
--- a/Snakefile
+++ b/Snakefile
@@ -35,6 +35,8 @@ rule merge_sces:
         directory(os.path.join(config["results_dir"], "merged_sce"))
     log:
         "logs/merge_sce.log"
+    params:
+        random_merge_flag = '--random_merge' if config['random_merge'] else ''
     shell:
         """
         Rscript scripts/02-prepare-merged-sce.R \
@@ -47,7 +49,7 @@ rule merge_sces:
           --num_hvg {config[num_hvg]} \
           --subset_hvg \
           --seed {config[seed]} \
-          {config[random_merge]} \
+          {params.random_merge_flag} \
           &> {log}
         """
 
@@ -83,6 +85,7 @@ rule integrate_fastmnn:
         "logs/{basedir}/{project}/integrate_fastmnn.log"
     params:
         merged_sce_file = "{project}_merged_sce.rds",
+        auto_merge_flag = '--fastmnn_auto_merge' if config['fastmnn_auto_merge'] else ''
     shell:
         """
         Rscript scripts/03a-integrate-sce.R \
@@ -91,7 +94,7 @@ rule integrate_fastmnn:
           --method fastMNN \
           --seed {config[seed]} \
           --corrected_only \
-          {config[fastmnn_auto_merge]} \
+          {params.auto_merge_flag} \
           &> {log}
         """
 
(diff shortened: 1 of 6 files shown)


config/config.yaml
sjspielman, line 1: If possible, I would set these up to be filled in _without_ dashes, as in `"random_merge"` and `"fastmnn_auto_merge"`. Then in the `Snakefile`, add the dashes back in, to be more consistent with other config flags we bring in.

...That said, of course, `snakemake` probably won't like that, which could be why you did it this way!

config/config.yaml
```diff
@@ -23,3 +23,7 @@ num_latent: 20
 # specify which projects to run through the workflow
 # to run a subset of projects use a python-formatted list '["1M_Immune_Cells", "HumanBrainSubstantiaNigra"]'
 groups_to_integrate: "All"
+
+# randomize merge order prior to integration
+random_merge: '' # --random_merge
+fastmnn_auto_merge: '' # --fastmnn_auto_merge
```
sjspielman, line 29: Can you also add a comment for `fastmnn_auto_merge`?

scripts/03a-integrate-sce.R
```diff
@@ -252,8 +261,19 @@ if (integration_method == "fastmnn") {
     batch_column = opt$batch_column,
     cosine_norm  = opt$fastmnn_no_cosine,
     gene_list    = fastmnn_gene_list,
-    seed         = opt$seed
+    seed         = opt$seed,
+    auto.merge   = opt$fastmnn_auto_merge
   )
+  
+  # add note about auto merge to
```
sjspielman, line 268: finish comment

scripts/02-prepare-merged-sce.R
```diff
@@ -21,6 +21,8 @@
 #   Only required if --add_celltype is used.
 # --batch_column: The name of the column in colData that indicates the batches for each cell,
 #   typically this corresponds to the library id. Default is "batch".
+# --random_merge: Used to indicate whether or not to merge SCE objects in a random order. 
+#   Default is FALSE.
```
sjspielman, line 25: Also need to add docs up here for new `seed` argument.

sjspielman (commented):
This all looks good to me, except it's missing the actual implementation of `fastmnn_auto_merge` in `scripts/utils/integrate-fastmnn.R` - need to add the argument, docs in this function, and add to `batchelor::fastMNN` itself. Once that is added here, I will go ahead and test the `Snakefile` on my end!

Something else just to note - I don't think that harmony order will matter here, but it'll be interesting if it does... If harmony order is moot, then we could do a different strategy of not during merging shuffling but just providing the argument `merge.order` to fastMNN, which is just a vector of the order to do it in, with `sample(length(combined_sce))`. But, we probably shouldn't do this since we need to definitely confirm whether order matters here for harmony. Just noting that concept here for posterity!
```
merge.order: An integer vector containing the linear merge order of
          batches in ‘...’. Alternatively, a list of lists representing
          a tree structure specifying a hierarchical merge order.
```

scripts/03a-integrate-sce.R
```diff
@@ -252,8 +261,19 @@ if (integration_method == "fastmnn") {
     batch_column = opt$batch_column,
     cosine_norm  = opt$fastmnn_no_cosine,
     gene_list    = fastmnn_gene_list,
-    seed         = opt$seed
+    seed         = opt$seed,
+    auto.merge   = opt$fastmnn_auto_merge
```
allyhawkins, line 265: I'm going to respond to other comments, but I just wanted to note that the auto merge is being implemented in `fastmnn` here. Because we have allowed for additional arguments already using `...` in `integrate_fastMNN()`, I'm using the `auto.merge` argument here without modifying the underlying scripts. I can do that if we think that's necessary, but I just wanted to note that the auto merge is being enabled here. 

https://github.com/AlexsLemonade/sc-data-integration/blob/2376f30227a7f57f288b5bc2fc90b0ca556398f7/scripts/utils/integrate-fastmnn.R#L30-L35

config/config.yaml
jashapiro: I think you could do this with values of `True` and `False` for the config, and then use `params` in the snakefile something like:
```python
params:
  auto_merge_flag =  lambda wildcards: '--fastmnn_auto_merge' if config['fastmnn_auto_merge'] else ''
```

(it may be that the `wildcards` above is not required, but I think it may be). 

followed by:
```
          --corrected_only \
          {params.auto_merge_flag} \
          &> {log}
```

I think `snakemake lint` will tell you not to use `config` values within the code anyway, but to convert them to `params` before use for better tracking.

scripts/03a-integrate-sce.R
allyhawkins: This is my preference too, since it leaves less room for error when updating in the future, which is why I didn't do it in the first place. I'm going to leave it out for now.

Snakefile
```diff
@@ -35,6 +35,8 @@ rule merge_sces:
         directory(os.path.join(config["results_dir"], "merged_sce"))
     log:
         "logs/merge_sce.log"
+    params:
+        random_merge_flag = '--random_merge' if config['random_merge'] else ''
```
jashapiro, line 39: oh that is nicer than having to use `lambda`

allyhawkins:
I went ahead and updated this to use the flags as true/false instead of directly having to put the flag in the config file. Thanks @jashapiro for the suggestion! I did test it with the new flags and everything worked as expected. This should be ready for another look.

scripts/03a-integrate-sce.R
sjspielman: > my preference would be to not to add more that we need to explicitly translate from one argument name to another.

makes sense!

sjspielman (approved):
🚀 pipeline runs all the way on my end!

PR #206, opened 2023-03-22
Add function for calculating reverse ARI

### LINKED ISSUE #202: Function for comparing pre correction clustering to post correction clustering
In research focus, we discussed the idea of testing a new integration metric to measure the correlation between clustering assignments before and after integration for a given set of samples. This metric is intended to identify if there is some over correction happening or all cells are being pushed together. To measure this we would want to calculate the ARI between the pre-correction and post-correction assignments. However, instead of calculating it for the entire merged SCE object, I believe we would want to obtain an ARI for each library or batch included in the integrated object. 

We will want to create a new function that is able to take in an integrated SCE object and a batch column and calculate the ARI between that and the integrated clustering results. Then for each batch it would get an ARI for only the assignments corresponding to that batch. I think we also will want to downsample and perform k-means clustering across a range of K for both the pre-integrated and integrated data and compare for each corresponding value and do that for each batch. 

This is fairly similar to the current `calculate_ari` function that we have, but we need to consider adding in the pre-integration clustering. If we can modify that function that would be great, but I don't want to risk messing up the accuracy of that function so it might be best to keep things separate for now as we test this out. 
The ultimate goal would be able to produce a similar plot to the batch/celltype ARI's that are currently in the summary notebook. 

We should see this section of [OSCA](http://bioconductor.org/books/3.16/OSCA.multisample/correction-diagnostics.html#preserving-biological-heterogeneity) for some guidance when tackling this.

allyhawkins:
Closes #202 

Here I added a new function for calculating the ARI between clustering pre and post integration. The function takes as input a list of individual SCE objects, before they go through merging, and the resulting integrated object containing the corrected PCs. For each individual SCE object and then integrated object, we grab the PCs and then go through downsampling and clustering using k-means. Then for each batch the ARI is computed between the clusters from the individual SCE object vs. the clusters from the integrated object that pertain to only the batch being compared. The result is a data frame with ARI, batch id, k, rep, and integration method. 

A few notes: 

- I wanted to be able to test with either scpca or simulated data. For the simulated data we didn't have pre-merged PCA that had been calculated so I made a small helper function that I added to just grab HVGs and calculate PCA for each provided SCE object if it doesn't exist already. If it does exist, it just grabs the PC matrix.
- For downsampling the integrated pcs, I had to do it slightly different here because after downsampling and clustering I need to subset to have a vector of the cluster assignments for only the batch id that we are interested in. This means that it needs to downsample the same fraction of cells across each batch in the integrated object, which our function didn't do. So I did some of my own data wrangling to downsample before looping through values of k. 
- I am slightly concerned about the actual ARI values that I'm getting, but I believe what I'm doing is the right principle based on the discussion in [OSCA](http://bioconductor.org/books/3.16/OSCA.multisample/correction-diagnostics.html#preserving-biological-heterogeneity). There they used the clusters obtained from each individual SCE object and compared to the clustering in the integrated object, looking at clusters for only that batch. However, aren't the cluster assignments themselves arbitrary for each object? As in the first cell listed will be in cluster 1 and then if the next cell is similar it will be in that same cluster or assigned to cluster 2 so comparing those labels directly seems error-prone. I'm getting a lot of ARI's around 0.01 which would make sense with arbitrary cluster assignments, but not if heterogeneity is preserved as they show in OSCA. I feel like I'm missing some sort of ordering or sorting step first, although I couldn't find that in any of the code in OSCA. Any ideas or thoughts on this would be helpful, to make sure I'm not missing something in the calculation. 
- I also added a few lines to the test script to test calculating this with the simulated data.


jashapiro:
I haven't gone into detail looking at your code, but I was confused in my first look: I would expect you to use the `bluster::pairwiseRand()` function to give a single value for each batch... something like
```
pairwiseRand(sce_list[[1]]$cluster, merged_sce[sce_merged$batch==1, ]$cluster, mode="index")
```

But applied to each batch, using batch labels and `purrr` most likely...

jashapiro:
> I haven't gone into detail looking at your code, but I was confused in my first look: I would expect you to use the `bluster::pairwiseRand()` function to give a single value for each batch... something like
> 
> ```
> pairwiseRand(sce_list[[1]]$cluster, merged_sce[sce_merged$batch==1, ]$cluster, mode="index")
> ```
> 
> But applied to each batch, using batch labels and `purrr` most likely...
> 
> The key point here, I think, is that there will be `n_batches` ARI values each time... which is different from the previous ARI where you were comparing a single set of batch labels for all samples.

Okay, looking at the code more, I think it is doing mostly what I would say, but it is a bit hard for me to track given the level of nesting of `purrr`. I think it would benefit from separating out some of the internal functions: One would take a single iteration of the clustering and calculate a table of ARIs for each batch. This should also make it easier to test if the ARI calcuation is doing what you expect (there should not be a need for the cluster labels to match, but that is easier to test with a separate function).

jashapiro:
One more thought: I think the k-means code here may be inappropriate/biasing against good comparisons. We don't expect the same number of clusters pre and post integration, so fixing K is probably not the best move here.  Consider B batches and C cell types, we expect BxC clusters before integration (assuming a big batch effect), and only C clusters after integration. So we might be better off with a single graph-based clustering for this metric.

allyhawkins:
Thank you @jashapiro! This was what I needed. I was trying to get ARI's for each batch, but with keeping the downsampling and testing different values of k that we had been previously using for the other ARI calculation, I think I was missing something. So I went ahead and did some major simplification for now, mostly to help with review, and to make sure that the calculation is correct. Then we can add in downsampling/ different k's later I think in a separate PR. 

I didn't quite do what you suggested by breaking it out into a completely separate function, because once I simplified things I think it got a lot clearer. This now returns a data frame with just the ARI and the batch ID. I also switched to using graph based clustering. Also these updates now give an ARI of 1 when looking at the simulated data. 🎉

scripts/utils/test_integration-functions.R
sjspielman, line 1: Might want to `file.path()`-ify (or `here::here()`) the paths in this file, for consistency

scripts/utils/calculate-ARI.R
```diff
@@ -92,3 +92,101 @@ calculate_ari <- function(integrated_sce,
...
+
+  
+  # get pcs for individual objects
+  ind_pcs <- purrr::map(individual_sce_list, get_individual_pcs)
+  
+  # Pull out the PCs or analogous reduction from integrated object
+  integrated_pcs <- reducedDim(integrated_sce, reduced_dim_name)
+  
+  # cluster integrated pcs only one time
+  integrated_clustering_result <- bluster::clusterRows(integrated_pcs,
```
sjspielman, line 157: With this level of nesting I'd suggest piping `integrated_pcs` into `bluster::clusterRows()` to make it a bit easier to read.

scripts/utils/calculate-ARI.R
```diff
@@ -92,3 +92,101 @@ calculate_ari <- function(integrated_sce,
...
+  ind_pcs <- purrr::map(individual_sce_list, get_individual_pcs)
+  
+  # Pull out the PCs or analogous reduction from integrated object
+  integrated_pcs <- reducedDim(integrated_sce, reduced_dim_name)
+  
+  # cluster integrated pcs only one time
+  integrated_clustering_result <- bluster::clusterRows(integrated_pcs, 
+                                                       bluster::NNGraphParam(cluster.fun = "louvain",
+                                                                             type = "jaccard",
+                                                                             k = k)) |>
```
sjspielman, line 160: `k` is not defined (could we just use the default of 10?)

> Error in h(simpleError(msg, call)) :
error in evaluating the argument 'BLUSPARAM' in selecting a method for function 'clusterRows': object 'k' not found

scripts/utils/calculate-ARI.R
```diff
@@ -92,3 +92,101 @@ calculate_ari <- function(integrated_sce,
...
+  integrated_clustering_result <- bluster::clusterRows(integrated_pcs, 
+                                                       bluster::NNGraphParam(cluster.fun = "louvain",
+                                                                             type = "jaccard",
+                                                                             k = k)) |> 
+    set_names(rownames(integrated_pcs)) # make sure to set the names with the batch ids
+  
+  
+  # for every batch id, cluster and then calculate ari for that batch 
+  all_ari <- batch_ids |> 
+    purrr::map_chr(\(batch){
```
sjspielman, line 166: Since ari is numeric, why return as `chr`? Maybe `map_dbl()` here?
```suggestion
    purrr::map_dbl(\(batch){
```

scripts/utils/calculate-ARI.R
```diff
@@ -92,3 +92,101 @@ calculate_ari <- function(integrated_sce,
...
+                                                                          k = k))
+      # extract clusters from integrated clustering for batch 
+      clusters_to_keep <- grep(batch, names(integrated_clustering_result))
+      batch_integrated_clusters <- integrated_clustering_result[clusters_to_keep]
+      
+      # calculate ari between pre-integrated clustering and post-integrated clustering for the given batch
+      ari <- bluster::pairwiseRand(ind_clustering_result, 
+                                   batch_integrated_clusters, 
+                                   mode = "index")
+
```
sjspielman, line 181: Also, TIL that anonymous functions return the most recent thing evaluated **even if it was assigned to a variable**, as is done here. 
So at the end here, let's put an explicit return here just to safe (but check spacing after accepting this suggestion!)

```suggestion
      return(ari)
```

scripts/utils/calculate-ARI.R
```diff
@@ -92,3 +92,101 @@ calculate_ari <- function(integrated_sce,
...
+      
+      # calculate ari between pre-integrated clustering and post-integrated clustering for the given batch
+      ari <- bluster::pairwiseRand(ind_clustering_result, 
+                                   batch_integrated_clusters, 
+                                   mode = "index")
+      
+    })
+  
+  # create data frame with ari and batch id
+  ari_df <- data.frame(
```
sjspielman, line 185: A personal preference of mine is I like making tibbles and not data frames because printing ;) (but up to you!!)
Also it makes it easier to see that `ari` is character when it should be numeric.

allyhawkins:
I went through all the comments and made most of the suggested changes with the exception of adding a `pc_name` argument to the new function to get a pca matrix from the individual objects. If there are strong feelings I can add that in, I just didn't think it was entirely necessary for the purpose of that function. I also updated the check so that this can now only be used when all batches in a merged object are present, so I added the rest of the batches into testing and everything still looks good in terms of calculating ARI. 

This is ready for another round of review.

scripts/utils/integration-helpers.R
```diff
@@ -577,4 +577,41 @@ remove_batch_nas_from_pcs <- function(integrated_pcs, batches) {
...
+#' @param sce SingleCellExperiment object
+#' @param num_genes Number of highly variable genes to select. Default is set to 2000.
+#'
+#' @return PC matrix
+
+get_individual_pcs <- function(sce,
+                               num_genes = 2000){
+  
+  # check that PC matrix is present in provided in SCE
+  if(!("PCA" %in% reducedDimNames(sce))){
```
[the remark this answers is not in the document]
sjspielman: Definitely fine with me! I do _not_ feel strongly.

scripts/utils/calculate-ARI.R
```diff
@@ -92,3 +92,100 @@ calculate_ari <- function(integrated_sce,
...
+                                  batch_column = "batch"){
+  set.seed(seed)
+  
+  # check that list of SCE objects is named
+  batch_ids <- names(individual_sce_list)
+  if(is.null(batch_ids)){
+    stop("Must provide a named list of SCE objects.")
+  } else {
+    # make sure the batch ids provided match between the list and the integrated object 
+    if(!(identical(sort(batch_ids), sort(unique(colData(integrated_sce)[, batch_column]))))){
```
sjspielman, line 128: Can we space this out a bit just for legibility as follows? I'm sure if I suggest it formally here I'll break the spacing in a worse way, RStudio automatic tabbing to the rescue!
```
if(!(identical(
  ...first one...,
  ...second one...
)
```

scripts/utils/calculate-ARI.R
```diff
@@ -92,3 +92,100 @@ calculate_ari <- function(integrated_sce,
...
+  
+  # get pcs for individual objects
+  ind_pcs <- purrr::map(individual_sce_list, get_individual_pcs)
+  
+  # Pull out the PCs or analogous reduction from integrated object
+  integrated_pcs <- reducedDim(integrated_sce, reduced_dim_name)
+  
+  # cluster integrated pcs only one time
+  integrated_clustering_result <- integrated_pcs |> 
+    bluster::clusterRows(bluster::NNGraphParam(cluster.fun = "louvain", type = "jaccard")) |>
```
sjspielman, line 158: Also maybe a new line here for legibility, like 
```
bluster::clusterRows(
   bluster::NNGraphParam(args)
)
```

scripts/utils/calculate-ARI.R
```diff
@@ -92,3 +92,100 @@ calculate_ari <- function(integrated_sce,
...
+      # extract clusters from integrated clustering for batch 
+      clusters_to_keep <- grep(batch, names(integrated_clustering_result))
+      batch_integrated_clusters <- integrated_clustering_result[clusters_to_keep]
+      
+      # calculate ari between pre-integrated clustering and post-integrated clustering for the given batch
+      ari <- bluster::pairwiseRand(ind_clustering_result, 
+                                   batch_integrated_clusters, 
+                                   mode = "index")
+      
+      ari
```
sjspielman, line 179: I'm ok with this, but I think some folks prefer - 

```suggestion
      return(ari)
```

scripts/utils/test_integration-functions.R
```diff
@@ -136,3 +136,25 @@ calculate_kbet(combined_sce, "batch", unintegrated=TRUE, seed = 2022,
...
+individual_sce_dir <- here::here("data", "scib_simulated", "sce")
+sce_ids <- c(
+  "sim1_Batch1",
+  "sim1_Batch2",
+  "sim1_Batch3",
+  "sim1_Batch4",
+  "sim1_Batch5",
+  "sim1_Batch6"
+)
+ind_sce_files <- glue::glue("{individual_sce_dir}/{sce_ids}_sce.rds")
```
sjspielman, line 150: check that my spacing doesn't end up wonky please, suggestions spacing is always eh... 

```suggestion
ind_sce_files <- file.path(
  individual_sce_dir, 
  glue::glue("{sce_ids}_sce.rds")
)
```

scripts/utils/test_integration-functions.R
```diff
@@ -136,3 +136,25 @@ calculate_kbet(combined_sce, "batch", unintegrated=TRUE, seed = 2022,
...
+individual_sce_list <- purrr::map(ind_sce_files, readr::read_rds) |>
```
sjspielman, line 151: #morepipesmorebetter

```suggestion
individual_sce_list <- ind_sce_files |>
  purrr::map(readr::read_rds) |> 
  purrr::set_names(sce_ids)
```

scripts/utils/calculate-ARI.R
```diff
@@ -92,3 +92,100 @@ calculate_ari <- function(integrated_sce,
...
+      
+    })
+  
+  # create tibble with ari and batch id
+  ari_tibble <- tibble::tibble(
+    ari = all_ari,
```
sjspielman, line 185: Just thought of this..should we call this `reverse_ari` in the end? Up to you!

```suggestion
    reverse_ari = all_ari,
```

sjspielman (approved):
Looks great, thanks! A couple super small spacing/path-related comments, and one comment about variable naming at your discretion. Don't need to see again!

scripts/utils/calculate-ARI.R
jashapiro: Maybe we should call this something less directional? `within_batch_ari` maybe? I don't really like "reverse ARI", as it is still ARI, just our expectation of the behavior is different.

sjspielman: +1 ^

PR #209, opened 2023-03-23
Remove immune and immune-related cell types from celldex refs for cell typing

### LINKED ISSUE #207: Create non-immune cell type refs from HPCA celldex ref
Related to #204, we want to be able to explore cell type assignments and have a control to look at. One of the controls we would like to explore, is a reference that does not contain immune cell types. All of the celldex references are immune cells or immune cells + other cells. One approach we thought of was to remove the immune cell types from the Human primary cell atlas reference so that we have one non-immune reference. This would also allow us to perform cell typing with the same reference with and without immune cells to see if the non-immune cells present in the dataset do in fact get correctly assigned to other cell types, or just get assigned because they are different from immune cells. 

Here I would create a script that grabs the HPCA reference from celldex, removes any immune cells, and then saves it as an `rds` file. We could then either add it to the metadata file used as input to the workflow in `scpca-nf` for training cell type refs, or perform the training and classification separately in a script in this repo.

allyhawkins:
Closes #207 

Here I'm beginning the section of this repo that we can use for exploring cell type annotations and adding a script that takes references we have from celldex and removes a list of celltypes, specifically immune cells. I started a new `celltype_annotation` folder where we should store any scripts or notebooks that are related specifically to exploring cell type annotation. I went back and forth on if this script should live with the other script that we have in `ScPCA-admin` to actually download the celldex refs and save them to S3 prior to them getting used as input to the workflows in `scpca-nf`. Ultimately I chose to file this here because I think we are still very much in an exploratory phase of trying to identify the best references and good control references to try. Because of that, I wanted to keep this here and if we want to end up using something like this for the actual workflow we can always move things later. 

The script here is designed to take as input a list of reference names to subset or remove cells from, e.g., `HumanPrimaryCellAtlasData`, and sync those references from where they live on S3 to a local `references` folder. It also reads in a text file that contains a list of celltypes or matching patterns to cell type names that should be removed from the reference object. To remove the celltypes it searches for anything matching that pattern in the `label.main` column of the `colData` and then removes those cells. Finally, a modified reference is saved to an rds file with an added prefix to the filename and then synced to the S3 bucket. 

Some thoughts I had/ reasons why I made some decisions: 

- We could directly grab the ref from `celldex`, but the reason I didn't is because I want to be working with the same version of the reference that we are using in `scpca-nf` that we have already downloaded. I don't think they should theoretically change, but just in case? 
- For removing cell types, the `label.main` is the broadest cell type so I thought that would be the easiest way to group and remove classes of cells. I don't foresee needing to get more granular at the time being, but I could make the label an argument if we really feel that is necessary. 
- To make the list of cell types/patterns, I went through both references I'm using right now and found all of the immune or blood related cell types. There were some that were the plural form so I only included the singular so that it would match regardless. There were some that included `HSC_CD34+` or `HSC_-G-CSF` so I added `HSC_` to the list to grab anything that started with that. There were a few instances like that I tried to grab. Unfortunately the T and B cells were annotated slightly differently between the two so I had to list them twice. 
- The reason I saved it back to S3 is because I want to be able to run our workflow in `scpca-nf` with these references so that we can produce objects that have annotations from both the original and the modified objects. To do that I will also need to add the filepath and name of the package now with the added prefix to the metadata that is pulled by the cell type workflow in `scpca-nf`. That will be the next follow-up PR. 

For hopefully easier reviewing, I'm going to attach the tables of `label.main` of each reference that I'm modifying so you can see which cell types we started with and where I got the names present in the list of cell types to remove: 

`BlueprintEncodeData`
Screen Shot 2023-03-23 at 4 34 04 PM

`HumanPrimaryCellAtlasData`
Screen Shot 2023-03-23 at 4 34 23 PM


celltype_annotation/immune_celltype_patterns.txt
```diff
@@ -0,0 +1,26 @@
...
+Neutrophils
+Pericytes
+B_cell
+BM
+CMP
+DC
+Embryonic_stem_cells
+Erythroblast
+GMP
+HSC_
```
jashapiro, line 19: Won't these be caught by `HSC`? above?

celltype_annotation/scripts/remove-cells-refs.R
```diff
@@ -0,0 +1,117 @@
...
+# create a regex from the list of celltypes/patterns to remove
+celltype_remove_pattern <- celltypes_to_remove |> 
+  paste(collapse = "|")
+
+# subset and remove desired cell types
+subset_ref_list <- purrr::walk2(ref_list, output_filepaths, 
+                                \(ref, output_file){
+                                  
+                                  # grab index of cells to remove 
+                                  removed_cell_idx <- grep(celltype_remove_pattern, ref$label.main)
```
jashapiro, line 102: For safety, do we want to use `stringr::str_starts()`? I assume that the patterns you are using are all prefixes, but correct me if I am wrong.

celltype_annotation/immune_celltype_patterns.txt
jashapiro, line 1: Can we alphabetize this list? It will make tracking later easier.

I was going to suggest standardizing on prefixes, but I'm now thinking we might want to be more specific to avoid worrying about false matches: we could make this two columns: reference & cell type, and then be specific about which ones we are removing. It will be a bit uglier but I think less likely to result in unexpected failures.

celltype_annotation/scripts/remove-cells-refs.R
jashapiro: Second thought: I think we should stick with exact matches on a per-reference basis, ultimately using `%in%`.

jashapiro (commented):
This looks like a good start, but I think for future maintenance, we may want to be specific about which labels apply to which reference, which means a table of the immune types, then pulling just those types for each reference to compare to. That will allow for exact matches, which I think is less likely to get us in trouble if we add a reference down the line.

Otherwise, I think the approach here looks fine. I agree with using the same reference file, and with sticking with `label.main`.

allyhawkins:
Based on the comments here I went ahead and created a table for the immune cell types and used the exact names for each of the cell types found in references. I then adjusted the code slightly to first filter the cell type table for the reference name and then use the `%in%` operator to find cells that should be removed. The table also has cell types alphabetized within each reference. This is ready for another look.

celltype_annotation/immune_celltypes.txt
```diff
@@ -0,0 +1,36 @@
...
+BlueprintEncodeData	Erythrocytes
+BlueprintEncodeData	HSC
+BlueprintEncodeData	Macrophages
+BlueprintEncodeData	Monocytes
+BlueprintEncodeData	NK cells
+BlueprintEncodeData	Neutrophils
+BlueprintEncodeData	Pericytes
+HumanPrimaryCellAtlasData	B_cell
+HumanPrimaryCellAtlasData	BM
+HumanPrimaryCellAtlasData	BM & Prog.
```
jashapiro, line 16: Oh my.

celltype_annotation/scripts/remove-cells-refs.R
```diff
@@ -0,0 +1,122 @@
...
+
+missing_ref_files <- local_ref_filepaths[!(which(file.exists(local_ref_filepaths)))]
+if(length(missing_ref_files)){
+  stop("Provided reference names are not associated with reference files either locally or in S3 directory.")
+}
+
+# read in ref
+ref_list <- local_ref_filepaths |> 
+  purrr::map(readr::read_rds) |>
+  setNames(opt$ref_names)
```
jashapiro, line 90: Why? tidyverse, I guess
```suggestion
  purrr::set_names(opt$ref_names)
```

celltype_annotation/scripts/remove-cells-refs.R
```diff
@@ -0,0 +1,122 @@
...
+
+# Remove cell types from refs --------------------------------------------------
+
+# create a regex from the list of celltypes/patterns to remove
+celltype_remove_pattern <- celltypes_to_remove |> 
+  paste(collapse = "|")
```
jashapiro, line 96: We're not using this anymore, are we?
```suggestion
```

celltype_annotation/scripts/remove-cells-refs.R
```diff
@@ -0,0 +1,122 @@
...
+               # subset to reference specific cell types
+               ref_specific_celltypes <- celltypes_to_remove |> 
+                 dplyr::filter(reference == ref_name)
+               
+               # vector of which cells to remove
+               removed_cells <- !(ref$label.main %in% ref_specific_celltypes$celltype)
+               
+               # subset reference and write out to new file
+               subset_ref <- ref[, removed_cells] |>
+                 readr::write_rds(output_file)
```
jashapiro, line 111: I flipped the logic here to match the variable names. I also added a `which`, which will protect us and keep any cells that might have `label.main` of `NA` (I don't think this exists, but seems worth doing)
```suggestion
               # vector of which cells to remove
               removed_cells <- which(ref$label.main %in% ref_specific_celltypes$celltype)
               
               # subset reference and write out to new file
               subset_ref <- ref[, -removed_cells] |>
                 readr::write_rds(output_file)
```

celltype_annotation/scripts/remove-cells-refs.R
```diff
@@ -0,0 +1,122 @@
...
+ref_filenames <- glue::glue("celldex-{opt$ref_names}.rds")
+  
+
+# output file names 
+output_filenames <- glue::glue("celldex-{opt$output_prefix}-{opt$ref_names}.rds")
+output_filepaths <- file.path(opt$ref_dir, output_filenames)
+
+# check that input list of cell types to remove has been provided 
+if(!file.exists(opt$celltypes_to_remove)){
+  stop("Must provide a txt file containing a list of cell types to remove.")
```
jashapiro, line 65:
```suggestion
  stop("Must provide a tsv file containing a table of cell types to remove.")
```

jashapiro (approved):
Looks good with the changes. I suggested a few little cleanups. I don't think I need to see it again.

PR #211, opened 2023-03-30
Plot within batch ARI

### LINKED ISSUE #203: Add a plot for reverse ARI and test with simulated data
After completing #202 and adding in the capability of calculating the ARI between clustering agreements for each batch in the integrated datasets, we will need to incorporate a plot to visualize the results into the template notebook. This should look very similar to the celltype/batch ARI plots that are there currently, where we have k on the x-axis and the ARI on the y axis and dots are colored by the batch. 

After that is added in, we should regenerate the reports with the simulated datasets prior to evaluating on any scpca data to make sure that the results look as expected.

allyhawkins:
Closes #203 

Here I added a plot for calculating the within batch ARI to the integration report. To be able to add this in I had to also add a way to first calculate the within batch ARI which requires having the individual SCE objects for all the libraries present in the integrated objects being evaluated. To do that I decided to add the library metadata file as a parameter. Then from there we can use the `integration_input_dir` column to define the directory that holds the original SCE objects and find them by filtering to the library IDs that are associated with the `params$group` being tested. I added the chunk that does the reading in of the individual files and creation of the list of individual SCE files towards the beginning of the notebook, after reading in the integrated data since it seemed to fit with that. Also I added a check to make sure all the expected files are present so I wanted that near the top. 

Then after looking at batch and celltype ARI, I added this new ARI where I just made one simple plot with integration method on the x-axis and ari on the y-axis. I didn't really think we needed anything too fancy here. I'm including two examples of testing this with the simulated data. 

Another thing I did here was remove the `unintegrated` option and actually allowed for `unintegrated` to be an `integration_method`. It makes being able to apply our function across the sce list that we have in the notebook a lot simpler. If people have strong methods about that I can change it back. I also didn't write a separate function here to calculate and then plot, because it wasn't super complicated. And having to return to this repo after a little break, I like having the code within the notebook rather than trying to parse through layers of functions. 

Simulated data with shared cell types
image

Simulated data where not all cell types are shared across all batches 
image

I'm going to tag both @jashapiro and @sjspielman and whoever gets to this first can take the review.

### DIFF
--- a/analysis_templates/01-single-group-integration-check-template.Rmd
+++ b/analysis_templates/01-single-group-integration-check-template.Rmd
@@ -174,7 +174,7 @@ library_ids <- stringr::str_remove(basename(individual_sce_files), "_sce.rds")
 
 # make sure that all the libraries we expect to be there have a corresponding file
 if(!all(library_ids %in% library_metadata$library_biomaterial_id)){
-  stop("Iindividual SCE objects not present for all libraries in group.")
+  stop("Individual SCE objects not present for all libraries in group.")
 }
 
 # create list of individual sce objects 
@@ -550,17 +550,17 @@ if (celltypes_present) {
 ```
 
 
-### Within batch ARI
+### Within-batch ARI
 
-Here we calculate the what we will refer to as the within bath ARI. 
-The goal of this metric is to measure how well a given integration method preserves biological heterogeneity by comparing clustering assignments from pre integration to post integration for each library. 
-This function takes each original SCE object, prior to any merging, and performs graph based clustering. 
+Here we calculate the what we will refer to as the within-batch ARI. 
+The metric measures how well a given integration method preserves biological heterogeneity by comparing pre- and post-intergration clustering assignments for each library. 
+This function takes each original SCE object, prior to any merging, and performs graph-based clustering. 
 The same clustering is then performed on the integrated object.
 For each library that is present in the integrated object, the original cluster assignments are compared to the new cluster assignments for those cells post integration by calculating the ARI between the cluster assignments. 
-Within batch ARI values close to 1 indicate that biological heterogeneity has been preserved. 
+Within-batch ARI values close to 1 indicate that biological heterogeneity has been preserved. 
 
 ```{r}
-# purrr across the list of integrated sce objects to compare against the individual objects 
+# map across the list of integrated sce objects to compare against the individual objects 
 within_batch_ari <- purrr::imap(sce_list, 
                                 \(sce, int_method){
                                   calculate_within_batch_ari(individual_sce_list,
--- a/analysis_templates/01-single-group-integration-check-template.Rmd
+++ b/analysis_templates/01-single-group-integration-check-template.Rmd
@@ -4,6 +4,7 @@ params:
   merged_sce_dir: "../results/scib_simulated/merged_sce"
   integrated_sce_dir: "../results/scib_simulated/integrated_sce"
   integration_methods: !r as.vector(c("fastmnn", "harmony", "cca", "rpca"))
+  library_metadata: "../sample-info/scib-simulated-processed-libraries.tsv"
   batch_column: "batch"
   celltype_column: "celltype"
   max_celltypes: 5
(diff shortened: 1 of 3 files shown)


analysis_templates/01-single-group-integration-check-template.Rmd
sjspielman, line 1: Leaving some initial comments here - At first I wondered why we can't just split up the unintegrated SCE object back into it's individual SCEs, rather than having to go back to each individual unintegrated file. My conclusion is this is because our merging _overwrites_ the PCs calculated from individual libraries, and those are the PCAs we want to use in these calculations. Just confirming here!

analysis_templates/01-single-group-integration-check-template.Rmd
```diff
@@ -508,6 +550,56 @@ if (celltypes_present) {
 ```
 
 
+### Within batch ARI
+
+Here we calculate the what we will refer to as the within bath ARI. 
+The goal of this metric is to measure how well a given integration method preserves biological heterogeneity by comparing clustering assignments from pre integration to post integration for each library. 
+This function takes each original SCE object, prior to any merging, and performs graph based clustering. 
+The same clustering is then performed on the integrated object.
+For each library that is present in the integrated object, the original cluster assignments are compared to the new cluster assignments for those cells post integration by calculating the ARI between the cluster assignments.
```
sjspielman, line 559:
```suggestion
The original cells' cluster assignments calculated on each individual library in the integrated object are then compared to the cells' new cluster assignments post-integration by comparing ARI between these two cluster assignments. 
```

sjspielman (commented):
Seems mostly good to me! I left some text suggestions, as well as one conceptual comment to confirm the approach here.

analysis_templates/01-single-group-integration-check-template.Rmd
sjspielman: Also, wanted to say - I think it would be helpful to add _small, short, very brief_ markdown sentence above each of these prep chunks that get hidden in the HTML. Something as quick as "Next, read in the unintegrated SCEs" above a chunk will make it easier to "unhide" given chunks of interest in the HTML.

allyhawkins: > Leaving some initial comments here - At first I wondered why we can't just split up the unintegrated SCE object back into it's individual SCEs, rather than having to go back to each individual unintegrated file. My conclusion is this is because our merging overwrites the PCs calculated from individual libraries, and those are the PCAs we want to use in these calculations. Just confirming here!

That is correct. We overwrite them when we calculate the batch corrected PC. Additionally, we subset the genes based on shared genes that are present in the merged object, so we couldn't go back and back-calculate the original PCs either.

allyhawkins:
I incorporated most of the suggestions that were made. As far as adding in small markdown comments in front of hidden code chunks, I added a few in front of the larger chunks that are more important but not all of them. I don't know if adding them to all the chunks is necessary for internal reports like this. This should be ready for another look.

analysis_templates/01-single-group-integration-check-template.Rmd
```diff
@@ -508,6 +554,56 @@ if (celltypes_present) {
...
+  # combine results into one dataframe
+  dplyr::bind_rows()
+```
+
+```{r}
+# set order of integration methods on axes
+within_batch_ari <- set_integration_order(within_batch_ari)
+
+# plot within batch ARI by integration method
+ggplot(within_batch_ari, aes_string(x = "integration_method_factor" , y = "ari")) +
```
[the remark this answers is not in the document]
jashapiro: > * for example:
> 
> ```
> a_real_function <- function(numeric_iris_column) {
>   ggplot(iris, aes(x = {{numeric_iris_column}})  + geom_histogram()
> }
> 
> # Usage is **without quotes**
> a_real_function(Sepal.Length)
> ```

I assume you meant something more like:

```
a_real_function <- function(data, numeric_column) {
  ggplot(data, aes(x = {{numeric_column}})  + geom_histogram()
}

a_real_function(iris, Sepal.Length)
```

Because I _know_ you would't use external variables in a function.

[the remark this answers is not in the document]
sjspielman: I want everyone to know, and this is true, that I wrote that example intentionally without `iris` to see if it would elicit this exact comment, so this is an all-around team success 🎉 🎉 

Of course, I would _never_ do such a thing in production 😄

sjspielman (approved):
🚀

PR #215, opened 2023-04-13
Explore marker genes based cell type assignment methods

### LINKED ISSUE #213: Test marker genes cell type annotation method
One thing we have noticed with cell type annotation using `SingleR` is that there regardless of the appropriateness of the reference, `SingleR` will always favor assigning a score rather than not labeling a cell. Although there are metrics that we have found may be helpful in determining if a reference is completely inappropriate (looking at the delta between the top score and median score), we would like to know if other types of cell annotation will be less likely to assign a cell type in general and will fail more easily. 

We should do a quick check of one of the marker genes based methods using the same Gawad libraries that we have been using in #210 and check the number of cells assigned to each type compared to using `SingleR`. 

The methods that we know that use marker genes are: 
- [cellassign](https://irrationone.github.io/cellassign/index.html)
- [scTyper](https://github.com/omicsCore/scTyper)
- [scType](https://github.com/IanevskiAleksandr/sc-type)

We should do this in an R notebook where we read in the annotated sce file that has been annotated using SingleR, assign cell types using one of the above methods, and then show the distributions of assignments. We can do a quick comparison of labels between the two methods where we are particularly interested in comparing how both methods do when using references that are inappropriate.

allyhawkins:
Closes #213 

Here I did some very preliminary analysis of two marker genes based cell type annotation methods. I want to emphasize the word preliminary here, as the goal of this was to answer the question of whether or not these methods are able to more quickly "fail" or assign `NA` to cells when provided an inappropriate reference vs. `SingleR`. We already know that `SingleR` has a preference to always assign cell types, while we hypothesized that these methods might lean towards not assigning cell types if no cells with those marker genes present could be identified. 

Also, some choices that I made were really just so that I spent less time on this, because ultimately all the methods except the `cellassign` python method are bad. 

In setting this up I tested 4 different methods and here's what happened: 

1. [`scType`](https://github.com/IanevskiAleksandr/sc-type): This method is available as functions in R that you can source from `Github` so dependencies are not managed very well and it's not actually an R package. I had to install `HGNChelper` and load the library before I could run the functions that I sourced in. They also have a database file that's an excel file with a list of marker genes for different groups of cells that you can use, or you can use your own. I definitely could have done some wrangling to get their marker genes into ensembl ids so we can use them with our SCE objects, but I had already spent way too much time on this. So I used the references we generated with `SingleR` and grabbed those marker genes as input. They also only have a function that produces a score matrix for each barcode, for each input celltype label so you get a score for every barcode for every label. Then the scores are summed for all cells in a cluster and the label associated with the top score is assigned to all cells in that cluster. The way they assign `NA` is by looking for cells that have a score that is 25% of the highest score or at least that was my interpretation. In summary, no `NA` cells were identified but also I don't really like how much I had to do to make things work and how much guessing I had to do around calculating scores and figuring out how to assign `NA`. 

2. [`scTyper`](https://github.com/omicsCore/scTyper):  I was able to get it installed but then I hit some of the errors that we've seen that aren't compatible with R 4.2 and I went back and looked and it uses R 3.5 so no thank you. 

3. [`cellassign` in R](https://irrationone.github.io/cellassign/index.html): This one requires `tensorflow` to be installed first and then you can install it. However, every time I installed `tensorflow`, I went to install `cellassign` and it couldn't find it. After spending too much time fighting with that, I found the python version from `scvi-tools` and moved on. 

4. [`cellassign` for `scvi-tools`](https://docs.scvi-tools.org/en/stable/tutorials/notebooks/cellassign_tutorial.html): I think if we're going to move forward with any of them, this is the one. This one was super easy to install and straight forward to use. There's a section in the notebook that I'm including where I output the annotated SCE as an hdf5 file and also modified the reference that I tested with to be compatible with our SCE object (switching out gene symbols to ensembl ids). Then there's a python script that performs annotation using `cellassign` and outputs the predictions matrix and the annotated hdf5 with two columns, one column for the reference with all cells and one column with the reference that removed B cells. Here I used the lymphoma reference that they had available on their [zenodo site](https://zenodo.org/record/3372746), because  you need a binary matrix with marker genes as rows and celltypes as columns. This one actually assigned some `other`, which means cells that didn't fit any of the input celltypes! 

Only `scType` and `cellassign` from `scvi-tools` are included in the PR.

Generally, I think the only method I would use moving forward, if any, was `cellassign`. It is slower obviously and would benefit from a GPU, so maybe something to keep in mind when moving towards implementing the `scverse`.  I also want to emphasize again that this is just a preliminary dive into these methods, so I'm sure things could be more efficient and cleaner, but in the effort of moving forward with this I was focused more on getting things to a point where we could clearly interpret the results. 

Here's a copy of the rendered notebook: 
[marker-gene-celltype-exploration.nb.html.zip](https://github.com/AlexsLemonade/sc-data-integration/files/11223850/marker-gene-celltype-exploration.nb.html.zip)


celltype_annotation/analysis/marker-gene-celltype-exploration.Rmd
```diff
@@ -0,0 +1,268 @@
...
+```
+
+
+```{r}
+# summary of assignment with no B cells 
+table(cellassign_annotated_sce$fl_noBcells_prediction)
+```
+
+`cellassign` actually will assign cells to `other` which is a column in the marker gene matrix representing cells that contain none of the marker genes. 
+We can also look at the actual predictions, similar to how we look at the scores in `SingleR`.
```
jashapiro, line 250: Yay! This seems very good, actually!

celltype_annotation/scripts/cell-assign.py
jashapiro, line 1: How long did this script actually take to run?

More importantly, how long did the `get_predictions()` function take? And was most of the time in the training or predicting?

 I think we can optimize quite a bit of other stuff: The conversion to anndata is going to be required to use this, but we don't actually have to convert back if we pull out the assignments only as a table (or using quarto & reticulate shared objects).

celltype_annotation/scripts/cell-assign.py
```diff
@@ -0,0 +1,65 @@
...
+
+# renormalize/ calcualate size factor on subset adata
+lib_size = annotated_adata.X.sum(1)
+subset_adata.obs["size_factor"] = lib_size / np.mean(lib_size)
+
+# small function for get cell type predictions given a marker genes matrix
+def get_predictions(anndata_object, marker_genes_mtx):
+    scvi.external.CellAssign.setup_anndata(anndata_object, size_factor_key="size_factor")
+    model = CellAssign(anndata_object, marker_genes_mtx)
+    model.train()
```
jashapiro, line 41: I'm a bit confused here... does the training have to use the `anndata_object`? Can we apply a single training to multiple datasets?

jashapiro (approved):
I think this is fine as an exploration document, but I would like to see just a bit more exploration of scType before we abandon it (maybe).

I think you may have misinterpreted the score cutoff, so I have a suggestion that I think matches the code in the README a bit better, and at least seems to make some kind of sense, even if it is entirely arbitrary.

I am also somewhat curious what the overall distribution of scores looks like at the individual cell level.  But maybe that is not worth doing, as I fully expect it to be very noisy. But we might well be able to do a more granular clustering (metaCellish). 

Cell assign looks pretty good though? I feel like we may well want to pursue that, but I'd be curious if the data here could include a bit more information about how long it takes to run. 

I feel like if we did want to use this we might well want to apply it with something like the `C8` dataset form [MSigDB](http://www.gsea-msigdb.org/gsea/msigdb/collections.jsp), but implementing that will be a whole other set of tasks, and it is absolutely fine not to include that here! 

A secondary thought: we should probably make `scverse` packages part of the `scpcaTools` docker image: that seems like it may be easiest way to manage that installation for use in workflows, especially if we want to use `reticulate` to handle interoperation between R and python.

celltype_annotation/analysis/marker-gene-celltype-exploration.Rmd
```diff
@@ -0,0 +1,268 @@
...
+    dplyr::add_count() |>
+    dplyr::top_n(n = 1, wt = sum_score)
+
+  # map everything back to each cell so that each cell receives a label
+  celltype_assignments <- coldata_df |>
+    dplyr::left_join(cluster_celltypes) |>
+    # get the score associated with the label assigned to that cell
+    dplyr::left_join(score_mtx) |>
+    # assign NAs for scores in the bottom 25% of all scores
+    dplyr::mutate(celltype = ifelse(score < max(score)/n, 'NA', celltype))
```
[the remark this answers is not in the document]
allyhawkins: Ah yes thank you, I missed the `$ncells` in their example.

celltype_annotation/analysis/marker-gene-celltype-exploration.Rmd
```diff
@@ -0,0 +1,268 @@
...
+  summarized_scores_mtx <- score_mtx |>
+    dplyr::group_by(clusters, celltype) |>
+    dplyr::summarise(sum_score = sum(score))
+  
+  # assign labels for each cluster, final number of rows should be equal to the number of clusters
+  # obtain one celltype label per cluster
+  cluster_celltypes <- summarized_scores_mtx |>
+    dplyr::group_by(clusters) |>
+    dplyr::add_count() |>
+    dplyr::top_n(n = 1, wt = sum_score)
```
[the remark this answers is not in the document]
allyhawkins: The matrix that is returned is a cell by label matrix so that you get a score for every cell for every label. Originally I looked at just assigning the label on a per cell basis by taking the label associated with the max score for each cell. I received similar results, so then tried the cluster method that was suggested. I can do both, but I wasn't quite sure how to assign `NA` for the per cell version since its based on the total number of cells per cluster. Maybe we use total number of cells in the dataset?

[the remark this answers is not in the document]
jashapiro: I think the score cutoff they are using is essentially `0.25*ncells` for a score that is the sum of the scores of individual cells, so maybe just using a score cutoff for a single cell of 0.25?

allyhawkins:
Thanks for the helpful comments @jashapiro! I did make some changes based on your comments in case you did want to take another look before merging: 

- For `scType`, I calculated a label on a per cell level and a per label level. In doing this I broke up the function I had a little, with one function for getting the score matrix and then two functions for assigning labels based on the method of choice. 
- I fixed the error in assigning `NA`, which actually did make a difference! There are more `NA` found when labeling by cluster, however this method is still not my favorite. Especially given the `CellAssign` results. 
- I also showed the scores that are assigned for `scType` and there is definitely a clear distinction between the reference that includes immune vs. not immune. We could probably incorporate similar metrics in regards to looking at the difference between the score and the median of scores, but I don't think this will provide any benefit over `SingleR`. 


- In regards to `CellAssign`, I removed adding the annotations directly to the ann data object and writing that object and re-reading it back in. Instead, I just save the prediction matrix and read that in to the R notebook. I did do some playing around with Quarto, but was having trouble using it with the conda environment we have and thought that at this stage it wasn't worth spending more time on doing that to get everything into one notebook. 
- I also checked the time to actually run the `get_predictions` function which includes training and prediction. The whole process is ~ 2 minutes with this test set of marker genes. 
- Additionally, it does appear we have to train for each library or dataset, as the `CellAssign` function expects an anndata object and the marker genes binary matrix. Based on what I looked at in the docs, I don't think we could separate them like we do with `SingleR`. 

As far as next steps, I think we would want to narrow in on some marker gene sets we would want to use (such as maybe the MsigDB C8 dataset or marker gene sets available on Azimuth or other public resources). Then we could test with a dataset that is more conducive to the samples we would be using and ensure we can replicate these results. 

Here's an updated report: 
[marker-gene-celltype-exploration.nb.html.zip](https://github.com/AlexsLemonade/sc-data-integration/files/11313858/marker-gene-celltype-exploration.nb.html.zip)

celltype_annotation/scripts/cell-assign.py
```diff
@@ -0,0 +1,57 @@
...
+# file path to annotated sce
+anndata_dir = os.path.join("..", "data", "anndata")
+annotated_adata_file = os.path.join(anndata_dir, "SCPCL000295_anndata.hdf5")
+annotated_adata = adata.read_h5ad(annotated_adata_file)
+
+# index adata but use row data to do it
+subset_adata = annotated_adata[:, fl_mtx.index].copy()
+subset_adata.X = subset_adata.X.tocsr()
+
+# renormalize/ calcualate size factor on subset adata
```
jashapiro, line 33: It looks like you are using the full data to calculate the size factors, but adding it to the subset, so I updated this comment. 

I wonder though if we can use the size factors we already calculated in R? Are those included in the `anndata` output somewhere? (Mostly a note for the future)
```suggestion
# add size factor to subset adata (calculated from full data)
```

celltype_annotation/analysis/marker-gene-celltype-exploration.Rmd
```diff
@@ -0,0 +1,344 @@
...
+```
+
+
+```sh
+# activate scvi environment
+conda activate scvi
+
+# create cell type assignments
+python ../scripts/cell-assign.py
+```
```
jashapiro, line 283: Just FYI, I'm working in the background on how `renv` works with python so that we might be able to better handle some of this installation & then use reticulate in the docker image. Hopefully.

jashapiro (approved):
> I also checked the time to actually run the get_predictions function which includes training and prediction. The whole process is ~ 2 minutes with this test set of marker genes.

That seems totally reasonable.

> Additionally, it does appear we have to train for each library or dataset, as the CellAssign function expects an anndata object and the marker genes binary matrix. Based on what I looked at in the docs, I don't think we could separate them like we do with SingleR.

Odd, but not a big deal given the fact that this only takes a couple of minutes.


Overall, this looks good to me! I commented on one potential refinement with size factors, but I think we can wait for implementation to work on the details.

PR #216, opened 2023-04-28
Render SingleR analysis reports for multiple libraries

### LINKED ISSUE #214: Evaluate metrics for SingleR on multiple libraries
After completing #204 and #210, we will have a template notebook for evaluating SingleR on a single library with annotated celltypes output from `scpca-nf`. Before proceeding with including the metrics that we would like to use in the cell type section of the QC report, such as the ARI between clusters and cell types and the delta median, we should render the report with multiple libraries. 

To start we should run the report on the Gawad libraries that we have annotated that are listed in the `scpca-processed-libraries.tsv` file. If we feel like we need to evaluate libraries from different projects we can go from there, but we should keep in mind that we would like to start with just datasets that are expected to have immune cell types.

allyhawkins:
Closes #214 

This PR adds a script to render the analysis template for looking at the SingleR output across a set of libraries. I used the existing metadata file that indicates which libraries we have processed and filter that based on the project we want to render reports for. Then for each library in that project, a report is rendered and saved using the template created in #210. 
The default is set up to render for the libraries in `SCPCP000007`, which should be 5 libraries that are currently in `scpca-processed-libraries.tsv`. 

When working on this, I also took some time to look at the results and was curious what the annotations looked like for the `label.fine` and `label.ont` label types, so I altered the plotting of the barplot to allow for plotting of all 3. I didn't make any other changes to the actual analysis in the report at this time.


celltype_annotation/scripts/render-singleR-report.R
```diff
@@ -0,0 +1,105 @@
...
+
+library_metadata <- readr::read_tsv(opt$library_metadata)
+
+if(!(opt$scpca_project_id %in% library_metadata$project_name)){
+  stop("Provided scpca_project_id not found in metadata file.")
+}
+
+# Render reports ---------------------------------------------------------------
+
+# filter to provided project ID and make sure column names match
```
jashapiro, line 94:
```suggestion
# filter to provided project ID and make sure column names match function arguments
```

celltype_annotation/analysis/SingleR-non-immune-ref-comparison.nb.html
jashapiro, line 1: This file doesn't seem to be a final render? It is missing the plots from the lower chunks (heatmaps and below).

jashapiro (approved):
This looks good to me, except for the fact that the example report file isn't fully rendered.

PR #219, opened 2023-05-19
Combine immune references for SingleR

### LINKED ISSUE #217: Test SingleR with combining immune references
As a follow up to the SingleR analysis, we should test how SingleR performs when combining into a single, harmonized reference. We should use the`label.ont` to harmonize references and follow the instructions in the [SingleR book](http://bioconductor.org/books/release/SingleRBook/using-multiple-references.html#using-harmonized-labels). 

This should probably be done in a separate analysis notebook as we will need to re-do the model training and classification.

allyhawkins:
Closes #217 

Here I did a quick test of looking at using `SingleR` with combining all the immune specific references in `celldex`. I used the [combining references section of SingleR book](http://bioconductor.org/books/release/SingleRBook/using-multiple-references.html#using-harmonized-labels) as a guide. I took the cell ontology labels from each reference and provided that to `SingleR`. This results in getting both an individual label for each reference included and then a single label that considers all references. We also obtain the information about which reference each label is from. 

I also did initially try assigning with just Blueprint and the HPCA dataset and it looked fairly similar with most of the cells just being assigned to the HSC group (CL:0000557). 

The one thing I did think about doing was looking at the score distributions, but I don't really know how to do that with the combined labels since there is only one score that gets output for each cell and then the rest is `NA`. I am definitely open to ideas on how we could validate that the combined reference is really "better" than the individual refs? 

I think my main goal of this was to see if using the combined reference may be a better approach to using individual references and picking one "good" reference. We talked about wanting to provide one single label to users from one reference that is most appropriate. Because of that I think I'm in favor of using the combined reference when we have a way of combining using ontology ID's, rather than trying to pick out which reference is the most appropriate for a given tissue. Given how `SingleR` will push to assign a cell type even if its not the most accurate, I'm in favor of including more options rather than using a limited reference. 
[SingleR-combining-refs.nb.html.zip](https://github.com/AlexsLemonade/sc-data-integration/files/11520412/SingleR-combining-refs.nb.html.zip)


celltype_annotation/analysis/SingleR-combining-refs.Rmd
```diff
@@ -0,0 +1,218 @@
...
+# read in all annotated sce 
+annotated_sce <- readr::read_rds(local_annotated_sce_path)
+```
+
+## SingleR with combined reference
+
+Before providing the list of references and labels, first subset them so that they contain the same set of genes.
+
+```{r}
+# identify shared genes among all references
```
jashapiro, line 87: From my read of the docs, it seemed like this may not really be a required step, but I doubt it will cause trouble.

celltype_annotation/analysis/SingleR-combining-refs.Rmd
```diff
@@ -0,0 +1,218 @@
...
+
+Before evaluating some of the output, I wanted to be able to interpret the ontology ID's so the first thing I did was map the `label.ont` back to the main labels from each reference then combine that with the SingleR results.
+
+```{r}
+# for each reference, provide a map between ont labels and other labels in the dataset
+ref_labels <- celldex_refs_shared |>
+  purrr::map(\(ref) data.frame(label_ont = ref$label.ont,
+                               label_main = ref$label.main,
+                               label_fine = ref$label.fine)) |>
+  dplyr::bind_rows(.id = "label_reference")
```
jashapiro, line 123: Alternatively:
```suggestion
  purrr::imap_df(\(ref, id){ 
    data.frame(
      label_reference = id,
      label_ont = ref$label.ont,
      label_main = ref$label.main,
      label_fine = ref$label.fine))
     )
   })
```

celltype_annotation/analysis/SingleR-combining-refs.Rmd
```diff
@@ -0,0 +1,218 @@
...
+# add main and fine labels to combined results 
+result_df <- combined_result |>
+  as.data.frame() |>
+  # first recode the reference assigned 
+  dplyr::mutate(reference = dplyr::case_when(reference == 1 ~ "BlueprintEncodeData", 
+                                             reference == 2 ~ "HumanPrimaryCellAtlasData",
+                                             reference == 3 ~ "DatabaseImmuneCellExpressionData",
+                                             reference == 4 ~ "MonacoImmuneData")) |>
+  # then combine with labels to get main and fine labels associated with ont labels for each reference
+  dplyr::left_join(ref_labels, by = c("pruned.labels" = "label_ont", "reference" = "label_reference"))
```
jashapiro, line 134: Doing it this way I think kind of defeats the purpose of using `label.ont`; we really only want one label per "cell type" and then use the same labels across all refs for visualization. The best way to do that is to look up in Cell Ontology what the preferred name is for each ontology

We should be able to that do that pretty easily (adapting a bit from [chapter 6])(http://bioconductor.org/books/release/SingleRBook/exploiting-the-cell-ontology.html#exploiting-the-cell-ontology)

```suggestion
# probably do this higher up and separately
cl_ont <- ontoProc::getOnto('cellOnto')

# add on cell ontology names
result_df |> 
  mutate(label_name = cl_ont$name[pruned.labels])
```

jashapiro (commented):
This looks good, it is still a bit tough where there may be cases of labels at different levels for "main" across ontologies. I would like to use the ontology labels as they come from Cell Ontology, and made a suggestion of how you can do that, rather than translating back to the varying reference labels. 

That might allow you to not just pull the top 10 labels, and might preserve some detail. 

I realized I'm also potentially concerned with how dplyr::join() may be behaving if there are multiple rows in the reference table values in a reference with the same `label.ont` values. (Is the output table the same size before and after the join?)

I would start with looking at the Cell Ontology name results, which you may also be able to use to get better labels in the plotScoreHeatmap.

jashapiro (commented):
This looks good, it is still a bit tough where there may be cases of labels at different levels for "main" across ontologies. I would like to use the ontology labels as they come from Cell Ontology, and made a suggestion of how you can do that, rather than translating back to the varying reference labels. 

That might allow you to not just pull the top 10 labels, and might preserve some detail. 

I realized I'm also potentially concerned with how dplyr::join() may be behaving if there are multiple rows in the reference table values in a reference with the same `label.ont` values. (Is the output table the same size before and after the join?)

I would start with looking at the Cell Ontology name results, which you may also be able to use to get better labels in the plotScoreHeatmap.

allyhawkins:
Okay I updated this to use the cell type names pulled directly from the cell ontology throughout. In doing that I actually removed the plot looking at all of the labels and just have the top 10. I can put it back or look at a higher number if you want, but it was a very busy plot and hard to interpret anything with so many labels and very long names for everything.

I also made the change from the very beginning which allowed me to use the labels in the heatmap too. The one thing I could not figure out was removing the annotation legend in the heatmaps and just show the row labels. 

Here's an updated report: 
[SingleR-combining-refs.nb.html.zip](https://github.com/AlexsLemonade/sc-data-integration/files/11569455/SingleR-combining-refs.nb.html.zip)

jashapiro:
> Okay I updated this to use the cell type names pulled directly from the cell ontology throughout. In doing that I actually removed the plot looking at all of the labels and just have the top 10. I can put it back or look at a higher number if you want, but it was a very busy plot and hard to interpret anything with so many labels and very long names for everything.
> 
> I also made the change from the very beginning which allowed me to use the labels in the heatmap too. The one thing I could not figure out was removing the annotation legend in the heatmaps and just show the row labels.
> 
> Here's an updated report: [SingleR-combining-refs.nb.html.zip](https://github.com/AlexsLemonade/sc-data-integration/files/11569455/SingleR-combining-refs.nb.html.zip)

It looks like there are changes here that might not have been pushed yet?

allyhawkins:
> It looks like there are changes here that might not have been pushed yet?

You're right, it would help if I had done that 🤦‍♀️  All good now!

celltype_annotation/analysis/SingleR-combining-refs.Rmd
```diff
@@ -0,0 +1,203 @@
...
+```
+
+It looks like the combined references look the most similar to the BlueprintEncodeData. 
+However, if we used individual references we see that the other references assign cells as CD34+ progenitors or monocytes. 
+In the combined reference we also see a small portion of the cells are annotated as CD34+ progenitors, which would not have been the case if we only used BlueprintEncodeData.
+
+```{r fig.height=20, fig.width=15}
+# make some heatmaps because why not
+SingleR::plotScoreHeatmap(combined_result,
+                          grid.vars = list(ncol = 1))
```
jashapiro, line 195: It is annoying that these heatmaps are all different sizes, which makes comparison among them difficult. I think we may be able to address this by using a set of labels found in all references, if they exist, and the `labels.use` parameter.

Some other parameters we might try from `pheatmap`, if you haven't already:

`annotation_legend = FALSE` to remove the legend.
We might also want to get rid of the row cluster tree with `treeheight_row = 0`

celltype_annotation/analysis/SingleR-combining-refs.Rmd
```diff
@@ -0,0 +1,203 @@
...
+  geom_bar(position = "stack", stat = "identity")  +
+  labs(x = "",
+       y = "Number of cells",
+       label = "Cell type") +
+  guides(x = guide_axis(angle = 90))
+  
+  
+```
+
+It looks like the combined references look the most similar to the BlueprintEncodeData.
```
jashapiro, line 188: This plot seems to me to be possibly the most useful one here. I am particularly interested by just comparing the blueprint to HPCA data, which seem to actually disagree on the labels "common myeloid progenitor" and "granulocyte monocyte progenitor" for a significant portion of cells. Blueprint "wins" here, but I honestly don't know what to make of that: is it a better reference, or just better matched to this dataset?

I think it might be interesting to plot a matrix of comparison between the the combined data and each reference, ideally some kind of annotation of the relationships among labels: when two references disagree, is there a parent/child relationship between the labels? We'd have to dive further into the ontology to create such a plot, so I think it is out of scope for this notebook. We would want to spend some time sketching out designs before trying to actually generate the plot, so this is a long-term thought.

jashapiro (approved):
This looks like it accomplishes the overall goal of having an example of the combined references, so I am approving it for now. I made some suggestions about the heatmap plot, but I don't think I need to see it again.

I do think this is something we might want to iterate on a bit if we are going to go down the multiple reference route. In particular, I gave some sketchy thoughts about how we might want to try to visually compare results from different references while accounting for ontological relationships. These are not well formed, but if we want to return to that it would be in a separate set of thoughts, focused on just the display question.

PR #221, opened 2023-06-07
Test Cell Assign against known cell types in sarcoma

### LINKED ISSUE #218: Try CellAssign on a different tissue type
Before we implement `CellAssign`, we would like to test it out to see how well it is able to assign cell types in a more diverse sample than an AML sample. We should pick another library from a more complex tissue (like the brain?) and run `CellAssign` on that tissue with the appropriate marker gene set. 

This will also require getting a list of marker genes for the cell types that we expect to find in the brain (or any other tissue that we test) and turning that into a binary matrix of genes x cell types. Some places that we found that we could grab a marker gene list: 
- [Azimuth](https://azimuth.hubmapconsortium.org/references/) 
- [PanglaoDB](https://panglaodb.se/)

allyhawkins:
Closes #218 

This PR does some additional testing regarding `CellAssign` to specifically look at how `CellAssign` performs in a non blood cell tissue type at assigning cell types. To address this, I started with a library from the Rhabdoymyosarcoma project where we have defined cell type annotations from the submitter. I wanted to compare the cell type annotations that we have to those assigned using `CellAssign` with two different marker gene lists. 

I used the following marker gene lists as references: 

1. [CellMarker2.0](http://117.50.127.228/CellMarker/CellMarker_download.html) - In looking at the methods of the manuscript, the marker list here was compiled based on manual literature searches. Here I used the `Human cell markers` file since it also contains the contents of the single cell only markers. The one thing I did like about this list is that we have cell ontology ID's for the cell types. 
2. [PanglaoDB](https://panglaodb.se/markers.html?cell_type=%27choose%27) - Here I downloaded using the `get tsv file` button which grabs the entire marker gene list. This list appears less comprehensive and we only have gene symbols. Also it hasn't been updated since 2020 so that's a downfall.

To use `CellAssign`, the marker gene lists need to be transformed into a gene by cell type binary matrix to indicate which genes belong to which cell type. I did this for both lists and then also added an "other" column where no genes are found. Any cells assigned to "other" would be the same as saying none of these cell types fit so we can't assign it. 

After running `CellAssign` with both references I compared the new cell annotations to the original using heatmaps. I did one where I grouped all tumor together and one where I broke up the tumor because I did notice that `CellAssign` was struggling with the tumor cells and does much better with normal cell types. This doesn't really surprise me though. With `CellMarker2.0` the tumor cells are assigned as macrophages though and none of those genes are really expressed in the dataset, so that does worry me. 

The last thing I did was definitely a bit circular, but my hope was that it would show me strong agreement between the assignments from `CellAssign` and original assignments. I used `scran::FindMarkers()` to identify the marker genes for each cell type assigned in our original dataset. Then I took the top 10 for each cell type to create the reference gene by cell type binary matrix. When I used that as a reference with `CellAssign` I saw much more consistent agreement between the cell type annotations, particularly with the normal cell types. However, I still noticed that `CellAssign` struggled with the tumor cell types and classified most of them as `other` when they should have been `Tumor_myoblast`. 

A few things I learned from this and where I think we should go next: 
1. `CellAssign` does like to not assign cell types when it doesn't think any of the cell types available are appropriate BUT this is still dependent on having a good quality reference. 
2. Classifying tumor cells is hard (should we be using CNV inference...) 
3. The time it takes to run `CellAssign` increases with the number of marker genes used/ size of the reference. In our initial testing we had ~ 5-10 genes and 4 cell types it took 2 minutes to run. These marker  gene matrices had up to 260 genes which took an hour to run 😬 I was running on just 1 cpu though so I would assume it gets faster as we increase that. 
4. No matter how great the method, if the reference is bad, the assignment is bad. I personally feel comfortable enough to continue with implementing `CellAssign` and `SingleR` with reporting two annotations. I think where we are going to need to do some tweaking is on what the appropriate references will be for each dataset. I also think the lists available on either CellMarker/ Panglao are good starting points but could be pared down and catered to each of our projects. Regardless of this... tumor cells are just going to be difficult... 

One small note is that when doing this I turned the `cell-assign.py` into a real script with arguments and changed how we run it in the previous notebook as well. I also have the python calls commented out in this notebook for rendering because it takes a long time to run locally. 

Here's the report: 
[cell-assign-sarcoma.nb.html.zip](https://github.com/AlexsLemonade/sc-data-integration/files/11679715/cell-assign-sarcoma.nb.html.zip)


celltype_annotation/analysis/cell-assign-sarcoma.Rmd
```diff
@@ -0,0 +1,435 @@
...
+```{r}
+# heatmap of prediction scores 
+cellmarker_predictions |>
+  dplyr::select(-barcode) |> 
+  as.matrix() |> 
+  pheatmap::pheatmap(show_rownames = FALSE)
+```
+
+Here most of the cells appear to be assigned to macrophages... 
+According to the known cell type annotations most of the cell types here are tumor cells and of those tumor cells they are mostly myoblasts.
```
jashapiro, line 200: Nothing appears to be assigned as a cancer type: are we sure they made it into the reference?
(the table I suggested above might be enhanced by translating the cell ontology labels)

celltype_annotation/analysis/cell-assign-sarcoma.Rmd
```diff
@@ -0,0 +1,435 @@
...
+  dplyr::filter(cell_name == "Macrophage")
+```
+
+What's strange is that CD163 and CD68 aren't present in our SCE object, only PTPRC and it is not highly expressed.
+
+```{r}
+# PTPRC is the only gene that's found in our dataset and the other two are not? 
+scater::plotReducedDim(processed_sce,
+                       dimred = "UMAP",
+                       colour_by = "ENSG00000081237")
```
jashapiro, line 214: Can you label the "macrophages" on this plot? I kind of doubt they are where this gene is expressed. It's all very odd.

celltype_annotation/analysis/cell-assign-sarcoma.Rmd
```diff
@@ -0,0 +1,435 @@
...
+
+```{r}
+# overview of available organs to choose from 
+unique(panglao_df$organ)
+```
+
+```{r}
+# grab all muscle related organs
+panglao_connective_tissue <- panglao_df |> 
+  dplyr::filter(organ %in% c("Connective tissue", "Smooth muscle", "Skeletal muscle"))
```
jashapiro, line 235: I'd probably like to see a table of the cell types and number of markers for each here as well.

jashapiro (commented):
Overall this looks good, and I appreciate the work that went into the various analyses you did. 

Many of my suggestions below are relatively minor code suggestions, with a view toward a bit  more reproducibility/easier updating. 

My other questions are a bit more about what the references look like before we try to use them for cell typing. How many markers are we using for some of these? Can that explain our lack of power? Making some tables of the cell types & numbers of markers from the marker DBs could be informative here, I think.

For the "circular" bit: how did you decide on 10 markers for each cell type? I also wonder if we might want to use "any" for these markers, or something in between: I kind of assume that most marker lists are some kind of a mix, where genes that mark a set of related types might be included in all of their marker lists. The fact that CellAssign uses a binary matrix certainly implies that it would expect some of that kind of data. (But does CellAssign require upregulation for a marker? I kind of assume so, but it is worth checking).

I might also move the retyping from submitter cell types to a different notebook: I feel like the exploration there is of a different character than the two published databases.

celltype_annotation/scripts/cell-assign.py
```diff
@@ -13,45 +13,68 @@
...
 # file path to annotated sce
-anndata_dir = os.path.join("..", "data", "anndata")
-annotated_adata_file = os.path.join(anndata_dir, "SCPCL000295_anndata.hdf5")
-annotated_adata = adata.read_h5ad(annotated_adata_file)
+annotated_adata = adata.read_h5ad(args.input_anndata)
 
 # index adata but use row data to do it
-subset_adata = annotated_adata[:, fl_mtx.index].copy()
+subset_adata = annotated_adata[:, ref_matrix.index].copy()
 subset_adata.X = subset_adata.X.tocsr()
```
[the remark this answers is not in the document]
allyhawkins: This unfortunately is required for scvi tools. Otherwise you get errors about not being the correct format. I followed the same code we use in https://github.com/AlexsLemonade/sc-data-integration/blob/3994b1934e255a129d5efcb56fa5fa2130d075dc/scripts/utils/integrate_scvi.py#L104-L108

celltype_annotation/analysis/cell-assign-sarcoma.Rmd
```diff
@@ -0,0 +1,435 @@
...
+  tidyr::pivot_wider('official gene symbol',
+                     names_from = 'cell type',
+                     values_from = 'cell type',
+                     values_fn = list('cell type' = length),
+                     values_fill = list('cell type' = 0)) |> 
+  tibble::column_to_rownames("official gene symbol") |>
+  dplyr::mutate(other = 0)
+
+# replace length with 1
+panglao_binary_mtx[panglao_binary_mtx > 1] <- 1
```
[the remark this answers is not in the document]
allyhawkins: I tried this and got a matrix that resulted in either 1 or `NA`, so I am sticking with the original way but removing the `list`.

allyhawkins:
> My other questions are a bit more about what the references look like before we try to use them for cell typing. How many markers are we using for some of these? Can that explain our lack of power? Making some tables of the cell types & numbers of markers from the marker DBs could be informative here, I think.

To address this I added tables to show the number of markers for each cell type in each of the databases. The number of markers seems to be as low as 1 and as high as 180, depending on the cell type and database. It looks like overall PangaloDB as more marker genes per cell type, at least for the muscle and connective tissues. 

> One further question: if you look at the `cellontology_id`s for the cancer types within muscle, do they show up anywhere else? As in, are we losing data by filtering to muscle first, when other datasets might give different markers for the same cell ontology id? This seems less likely, but it could be a concern.


I also added tables that might clarify the inclusion of the cancer types in the marker gene database for CellMarker. I looked at all the cancer types that contain the word "sarcoma" and made a table of the cell types and associated marker genes. We do see that some of the cell types are duplicated across cancers and have different numbers of marker genes. This tells me that the marker genes used for a myoblast in Osteosarcoma are not the same as in RMS, which concerns me.

> Can you label the "macrophages" on this plot? I kind of doubt they are where this gene is expressed. It's all very odd.

I updated the plot showing the expression of the macrophage marker genes to be split by cell type. In addition to that, I chose to look at some of the marker genes for Myoblasts, since that cell type is also included in the CellMarker reference we used but no cells are assigned to it. I took a look at the paper with the original RMS data and they had pulled out `MSC`, `MYF5`, and `PAX7` as myoblast markers in their dataset. These markers are also included in the marker gene list for myoblasts from CellMarker. I took a look at the expression of those genes and noticed that the expression seemed fairly low in this particular sample. 
Maybe we should consider running this on other samples to see how consistent these results are. 

> I might also move the retyping from submitter cell types to a different notebook: I feel like the exploration there is of a different character than the two published databases.

I did break this up into two notebooks and I'm going to file a separate PR for using the marker genes as a reference. In doing this, I needed to break out the functions for getting the cell assignments from the CellAssign predictions to a separate file I can source into both notebooks. 

Attached is an updated version of the notebook: 
[cell-assign-sarcoma.nb.html.zip](https://github.com/AlexsLemonade/sc-data-integration/files/11840660/cell-assign-sarcoma.nb.html.zip)

celltype_annotation/analysis/cell-assign-sarcoma.Rmd
```diff
@@ -0,0 +1,451 @@
...
+# read in excel file
+celltype_markers_df <- openxlsx::read.xlsx(excel_celltype_file)
+```
+
+
+```{r}
+# look at different tissue types available 
+celltype_markers_df |>
+  dplyr::count(tissue_class)
+```
```
jashapiro, line 97: I'm a bit confused about the organization/origins of this data, so I think a little explanatory paragraph at the start would be useful. To the extent that you are able to find that information, that is.

Mostly I am a a little confused about the interaction between cancer type and tissue type: is every set of markers divided into tissues and which may or may not include cancer types?

celltype_annotation/analysis/cell-assign-sarcoma.Rmd
```diff
@@ -0,0 +1,451 @@
...
+
+```{r}
+# grab the muscle tissue type
+muscle_markers_df <- celltype_markers_df |> 
+  dplyr::filter(tissue_class == "Muscle") |>
+  tidyr::drop_na(cellontology_id, Symbol)
+
+# print total number of marker genes for each cellontology id
+muscle_markers_df |>
+  dplyr::count(cell_name, cancer_type, cellontology_id)
```
jashapiro, line 115: Okay, so what does it mean to divide a cell type between cancer types here? Is the single marker of myoblasts in rhabdomyosarcoma a subset of the 35 in normal tissue? Or is it in addition to the normal markers? Or is it from a different set of experiments? (Same question for stem cells and presumably every other cell type, but that was the first one that really jumped out to me)

The answers to these questions may not all be important, but might affect if/how we would choose to use this resource.

(Also confused by skeletal muscle not showing up in normal)

celltype_annotation/analysis/cell-assign-sarcoma.Rmd
```diff
@@ -0,0 +1,451 @@
...
+`MSC` is not in this list and not expressed in our dataset. 
+The other two are both in the marker gene list and in our dataset, but how highly are they expressed in the population that gets labeled as `Tumor_myoblast`? 
+
+```{r}
+# marker genes for myoblasts from paper
+# MYF5 ENSG00000111049
+# MSC ENSG00000178860 (not in marker genes listed from cellmarker or our dataset)
+# PAX7 ENSG00000009709
+
+scater::plotReducedDim(processed_sce,
```
jashapiro, line 263: I wonder if `scater::plotExpression` might be better here.

celltype_annotation/analysis/cell-assign-sarcoma.Rmd
```diff
@@ -0,0 +1,451 @@
...
+panglao_connective_tissue |>
+  dplyr::count(`cell type`, `organ`)
+```
+
+
+```{r}
+# create a binary matrix of genes by cell type markers 
+panglao_binary_mtx <- panglao_connective_tissue |>
+  dplyr::select('cell type', 'official gene symbol') |> 
+  tidyr::pivot_wider('official gene symbol',
```
jashapiro, line 308:
```suggestion
  tidyr::pivot_wider(id_cols = 'official gene symbol',
```

jashapiro (commented):
This looks pretty good! By which I mean I am very confused, but mostly by the results, not by what you have written. I like the comparisons at the end of the "expert" assignments with the automated methods.

I guess my main conclusion (you might add yours to the end?) seems to be that panglao may be a more useful database of markers for at least this sample... I really don't like that we seem to get assignments based on only a single marker. Though now I wonder if that is happening? When you pivot by `cellontology_id`, that combines the "cancer" and "normal" cells, doesn't it? Do we get different result if we filter _out_ the markers that only show up in the cancer cell types? Naively, I would expect those to be less reliable, at least for now. 

I also worry a bit, as you mentioned, about markers that differ depending on the "enclosing" tissue type. Does that also seem to be the case in panglao? It seems like there the cell types are each assigned to only a single tissue type, at least in the examples I could see in the notebook. (Of course, these don't have cell ontology ids, which is their own limitation).

Overall, I don't think there are any major changes to the code needed here, aside from maybe the plots (I probably suggested the UMAPs, but shouldn't have).

The main addition I'd still like to see is a bit more explanatory text for when we come back to these to figure our our decision process.

allyhawkins:
I made a few more changes based on your most recent comments so this is ready for another round of review. 

- I provided detailed information on the marker gene lists for each database, including how they were created and organized. Or at least I filled in as much as I could gather from the websites and papers. 
- Instead of UMAPs, I utilized `scater::plotExpression` to display the gene expression of marker genes among different cell types.
- I added a conclusion section with some general comments on the findings. I, too am worried about the macrophage assignment and find it a little concerning. If we combined all the macrophage markers from all tissue types in CellMarker2.0, would `CellAssign` do a better job? But I worry we would get the same result as we did with PanglaoDB, where most of the cells are designated as "other". 

Attached is the updated notebook: 
[cell-assign-sarcoma.nb.html.zip](https://github.com/AlexsLemonade/sc-data-integration/files/11884342/cell-assign-sarcoma.nb.html.zip)

celltype_annotation/utils/cellassign-helper-functions.R
```diff
@@ -0,0 +1,32 @@
...
+# define a function for obtaining cell type assignments from cell assign predictions
+get_celltype_assignments <- function(predictions){
+  
+  # get a dataframe of assignment for each barcode and prediction score
+  celltype_assignments <- predictions |>
+    tidyr::pivot_longer(!barcode,
+                        names_to = "celltype",
+                        values_to = "prediction") |>
+    dplyr::group_by(barcode) |>
+    dplyr::top_n(n = 1, wt = prediction) |>
```
jashapiro, line 11: Apparently `top_n` is deprecated
```suggestion
    dplyr::slice_max(prediction, n = 1) |>
```

celltype_annotation/analysis/cell-assign-sarcoma.Rmd
```diff
@@ -0,0 +1,481 @@
...
+```{r}
+# print out markers that were associated with macrophages during assignment
+muscle_markers_df |>
+  dplyr::filter(cell_name == "Macrophage")
+```
+
+What's strange is that CD163 and CD68 aren't present in our SCE object, only PTPRC and it is not highly expressed.
+
+```{r}
+processed_sce$cellmarker_assignments <- cellmarker_assignments$celltype
```
jashapiro, line 242: Worth checking that the barcodes are in the same order here before adding the annotations to the SCE object

celltype_annotation/utils/cellassign-helper-functions.R
```diff
@@ -0,0 +1,32 @@
+
+# define a function for obtaining cell type assignments from cell assign predictions
+get_celltype_assignments <- function(predictions){
+  
+  # get a dataframe of assignment for each barcode and prediction score
+  celltype_assignments <- predictions |>
+    tidyr::pivot_longer(!barcode,
+                        names_to = "celltype",
+                        values_to = "prediction") |>
+    dplyr::group_by(barcode) |>
```
jashapiro, line 10: I worry that the `group_by` may be altering order here.

jashapiro (commented):
Just a quick comment here to note places where order might be checked.

Not a full review!

celltype_annotation/analysis/cell-assign-sarcoma.Rmd
```diff
@@ -0,0 +1,481 @@
...
+
+This means that there are duplicates of the same cell type that are found in different tissue types.
+There also will be duplicates within a tissue type if a cell type is found in both a normal and cancer sample.
+
+For example, macrophages are found in multiple tissue types so the marker gene list for macrophages may change based on the tissue type you look at.
+Additionally, macrophages are in both normal and cancer cells so a different marker gene list for macrophages in normal muscle will vary from macrophages in cancer tissue from muscle.
+All cell types that are in normal tissue are not always listed in cancer tissue and vice versa.
+
+Because we are creating a gene by cell type matrix using the ontology ID and Ensembl ID, all marker genes for a given ontology ID will be combined into a single list of marker genes for that cell type.
+However, if we filter on tissue type first and then create this matrix, then we may miss marker genes used for a given cell type that is also found in another tissue type.
```
jashapiro, line 103: I wonder if this is something we can try to work around? Can we use the tissue type to get our cell type list, but then use _all_ markers for that cell type regardless of tissue? That seems potentially dicey, but also not crazy to me. I don't know how to evaluate it, except by feel.

celltype_annotation/analysis/cell-assign-sarcoma.Rmd
```diff
@@ -0,0 +1,481 @@
...
+As expected, `CellAssign` seems to be struggling with the tumor cells and depending on the reference has been assigning them to `other` or `macrophage`. 
+There are a few places where the normal cells match up in their assignment, but they are also not perfectly in agreement. 
+
+## Conclusions
+
+- Looking at the two marker gene lists, PanglaoDb appears to have a larger number of marker gene lists for each cell type than CellMarker2.0.
+This is probably due to the way the marker gene list is organized.
+PanglaoDb only provides one list of potential marker genes for each cell type, regardless of tissue type. 
+- CellMarker2.0 has a larger list of tissue types and cell types and even includes some cancer tissue, while PanglaoDb does not. 
+Additionally, it provides the added advantage of grouping marker genes by ontology ID.
```
jashapiro, line 467: The fact that CellMarker uses cell ontology is probably its biggest benefit at this stage. PanglaoDb not using it is annoying, but maybe not awful given that we don't really trust anything at this point.

celltype_annotation/analysis/cell-assign-sarcoma.Rmd
jashapiro, line 242: One way to make sure the order matches:
```suggestion
cellmarker_assignments <- data.frame(barcode = processed_sce$barcode) |>
  dplyr::left_join(cellmarker_assignments, by = barcode)
  
processed_sce$cellmarker_assignments <- cellmarker_assignments$celltype
```

jashapiro (commented):
Thanks for adding the comments about the references, as I think they are quite helpful!

Once we check for potential reordering in the cell barcode list, I think we should be good to go.

allyhawkins:
I checked the order before adding the cell type assignments to the `processed_sce` and re-ran the notebook. We are still getting the same results, so unfortunately, I don't think that was it. 

> I wonder if this is something we can try to work around? Can we use the tissue type to get our cell type list, but then use _all_ markers for that cell type regardless of tissue? That seems potentially dicey, but also not crazy to me. I don't know how to evaluate it, except by feel.

I don't think this idea is that wild, but I'm unsure how we would evaluate it (like you said). I think I like the approach of combining organs from PanglaoDb rather than altering the marker gene lists. I don't have a strong rationale, but not altering the lists seems safer, considering we don't fully understand how they were created. We would lose out on the ontology ID benefit of CellMarker, which is unfortunate.

jashapiro:
> We would lose out on the ontology ID benefit of CellMarker, which is unfortunate.

I actually think we can probably do a pretty good mapping to cell ontology if we want to do this. We might have to do some manual checking, but I bet we can make much of it pretty automatic with a little effort

PR #223, opened 2023-06-22
Use marker genes from defined cell types as reference for CellAssign

allyhawkins:
Stacked on #221 

This PR is a continuation of the analysis started in #221. Here I broke out the comparison of using the top marker genes as a reference for cell assignment as a separate notebook. Before filing this, I addressed some of the concerns brought up in the initial review: 

> For the "circular" bit: how did you decide on 10 markers for each cell type? I also wonder if we might want to use "any" for these markers, or something in between: I kind of assume that most marker lists are some kind of a mix, where genes that mark a set of related types might be included in all of their marker lists. The fact that CellAssign uses a binary matrix certainly implies that it would expect some of that kind of data. (But does CellAssign require upregulation for a marker? I kind of assume so, but it is worth checking).

- Honestly, I don't really have a rationale for picking 10 markers. I just liked the number and was trying to grab the most highly expressed marker genes for each cell type without getting too many genes that may limit the ability to differentiate cell types. 
- I didn't change the number of marker genes used, but I did slightly change how I grab them. Previously I wasn't doing any sorting based on log fold change. Here before creating the matrix, I only grab the top 10 markers with the highest positive fold change. 
- I did read a bit more about CellAssign to confirm that the expectation of the marker genes is that they are highly expressed in a given cell type in comparison to all other cell types. See some of the text taken directly from the paper https://www.nature.com/articles/s41592-019-0529-1. 
	- "CellAssign allows for flexible expression of marker genes, assuming that marker genes are more highly expressed in the cell types they define relative to others."
	- "Marker genes are assumed to be overexpressed in cell types where they are markers—not necessarily at similar levels—compared to those where they are not."
- I looked at the top markers identified for the `Tumor_myoblast` population and of the top 10 (I also looked at the top 20 on my own) none of them were genes that were mentioned in the paper. Also, none of them were Myoblast specific genes that I recognized (`MYOG`, `MYOD1`, `MYF5`). I looked a few up and most of them seemed more ubiquitously expressed. 
- Despite none of them being myoblast-specific genes, using them to define Myoblasts was able to recapitulate cell type assignments fairly well. Using only the markers with the highest fold change meant the assignments were more consistent with the manual annotations, which is what I expect. 

Here's a copy of the completed notebook: 
[cell-assign-sarcoma-marker-genes.nb.html.zip](https://github.com/AlexsLemonade/sc-data-integration/files/11840682/cell-assign-sarcoma-marker-genes.nb.html.zip)


celltype_annotation/analysis/cell-assign-sarcoma-marker-genes.Rmd
```diff
@@ -0,0 +1,194 @@
...
+processed_sce <- readr::read_rds(local_processed_sce_path)
+```
+
+```{r}
+# source helper functions
+function_file <- file.path("..", "utils", "cellassign-helper-functions.R")
+source(function_file)
+```
+
+## Get marker genes
```
jashapiro, line 59: I would add just a bit of text here to say what the marker genes you are finding are marking, and where the cell types came from.

celltype_annotation/analysis/cell-assign-sarcoma-marker-genes.Rmd
```diff
@@ -0,0 +1,194 @@
...
+
+
+```{r}
+# get top 20 markers for each cell type 
+top_markers_df <- as.list(processed_sce_markers) |>
+  purrr::imap(
+    \(markers, id) 
+    markers |> 
+      as.data.frame() |>
+      dplyr::arrange(desc(summary.logFC)) |>
```
jashapiro, line 88: Probably should comment that we are only looking "upregulated" genes here, because that is what CellAssign needs (I think that is what we discussed).

celltype_annotation/analysis/cell-assign-sarcoma-marker-genes.Rmd
```diff
@@ -0,0 +1,194 @@
...
+  dplyr::select(barcode, celltype) |>
+  # classify cells as tumor or normal
+  dplyr::mutate(cell_category = dplyr::if_else(stringr::str_detect(celltype, "Tumor"), "Tumor", celltype),
+                # remove sub states of myoblast/mesoderm/myocyte
+                broad_celltype = stringr::str_remove(celltype, "-A|-B|-C|-D"))
+```
+
+
+```{r}
+# find markers for all cell types
```
jashapiro, line 74: Worth noting that you are looking for markers that are "unique" to each cell type. (This may be a reason for some uncertainty: there is no ability to do any kind of "hierarchical" model. 

(This has me doing some deeper thinking about how best to select markers automatically, but I don't have any answers, and this is not where to explore it!)

celltype_annotation/analysis/cell-assign-sarcoma-marker-genes.Rmd
```diff
@@ -0,0 +1,194 @@
...
+top_markers_df <- as.list(processed_sce_markers) |>
+  purrr::imap(
+    \(markers, id) 
+    markers |> 
+      as.data.frame() |>
+      dplyr::arrange(desc(summary.logFC)) |>
+      dplyr::mutate(celltype = id) |>
+      tibble::rownames_to_column("ensembl_id") |> 
+      dplyr::select(celltype, ensembl_id) |>
+      head(n = 10)
```
jashapiro, line 92: Comment above says top 20. 

I don't know how we would do if we used a logFC cutoff? Again, not required for this analysis, to be sure.

celltype_annotation/analysis/cell-assign-sarcoma-marker-genes.Rmd
```diff
@@ -0,0 +1,194 @@
...
+    org.Hs.eg.db::org.Hs.eg.db,
+    # the provided gene identifiers
+    keys = myoblast_markers$ensembl_id,
+    # the type of provided gene identifiers
+    keytype = "ENSEMBL",
+    # the type of gene identifiers to map to
+    column = "SYMBOL"
+  ) |>
+    tibble::enframe(name = 'ensembl_id',
+                    value = tolower("SYMBOL")) |>
```
jashapiro, line 133: Any reason not to just give the column name you want here?
```suggestion
                    value = 'symbol') |> 
```

celltype_annotation/analysis/cell-assign-sarcoma-marker-genes.Rmd
```diff
@@ -0,0 +1,194 @@
...
+    # the type of gene identifiers to map to
+    column = "SYMBOL"
+  ) |>
+    tibble::enframe(name = 'ensembl_id',
+                    value = tolower("SYMBOL")) |> 
+    dplyr::left_join(myoblast_markers, by = "ensembl_id")
+
+myoblast_markers
+```
+I kind of would expect some of the myoblast markers to be included here like `MYOG`, `MYF5`, `MYOD1`, etc.
```
jashapiro, line 138: Are they not present because those would also be present in non-tumor myoblast cells? Are those among the cells annotated here? If so, we might expect to recover those if using `pval.type = "any"` in `findMarkers()`.

(My thought earlier was to use some combination of the markers found by `any` and `all` to supply to CellAssign)

celltype_annotation/analysis/cell-assign-sarcoma-marker-genes.Rmd
```diff
@@ -0,0 +1,194 @@
...
+
+```{r}
+# heatmap of prediction scores 
+rms_predictions |>
+  dplyr::select(-barcode) |> 
+  as.matrix() |> 
+  pheatmap::pheatmap(show_rownames = FALSE)
+```
+
+This actuallylooks pretty good!
```
jashapiro, line 173:
```suggestion
This actually looks pretty good! 
```

jashapiro (commented):
I think this looks good as a preliminary/supplementary analysis. I had a thought for where it might go next, but I don't think you need to implement any of that now, if at all. I would like to see just a bit more background text about how markers are being chosen though!

allyhawkins:
Thanks for taking a look at this @jashapiro. I added some explanatory text around the analysis that was already here, but I didn't update any plots or any of the actual analysis. 
I will note that I did try to use `any` rather than `all` for the `FindMarkers` strategy and there was not a noticeable difference which is why I just kept using `all`. I do think the idea of doing a logFC cutoff is good, but I'm not sure we need it at this stage.

jashapiro (approved):
Seems good for this stage!

Issue #224, opened 2023-06-28
Rerun SingleR analysis with RMS libraries


allyhawkins:
We would like to evaluate using `SingleR` to annotate cell types on a solid tumor type. We already have a template R notebook, `SingleR-non-immune-ref-comparison.Rmd`, which we used to evaluate `SingleR` on libraries from a liquid tumor type. We just need to render that notebook with a few of the libraries from the RMS project.

PR #227, opened 2023-07-11
Minor updates to rendering SingleR report

allyhawkins:
When working on #224 I made a few minor changes to the script I had for rendering the SingleR analysis. 
- Removed an extra slash in the results directory 
- Update the output of the rendered reports to be nested by project ID rather than outputting all the reports to a single directory. This made it easier to find the reports that were for the project that was being run rather than trying to match library ids to project ids.


celltype_annotation/scripts/render-singleR-report.R
```diff
@@ -39,7 +39,7 @@ option_list <- list(
   make_option(
     opt_str = c("--s3_prefix"),
     type = "character",
-    default = "s3://nextflow-ccdl-results/scpca/processed/results/",
+    default = "s3://nextflow-ccdl-results/scpca/processed/results",
```
sjspielman, line 42: 🎢

celltype_annotation/scripts/render-singleR-report.R
sjspielman, line 1: I can't comment on this directly, and it doesn't need to change since it works fine.. Just FYI! `dplyr::select()` lines `98-100` - you don't need quotes around new column names (e.g. `"project_id" = project_name` could just as well be `project_id = project_name`) because `tidy-select` is 🪄

sjspielman (approved):
👍

celltype_annotation/scripts/render-singleR-report.R
jashapiro: Just to continue the rollercoaster, I would leave this change, but you should probably not use `file.path()` at https://github.com/AlexsLemonade/sc-data-integration/blob/9e3f10fc94bf6c6da8144622367fef423c5a4b6b/celltype_annotation/scripts/render-singleR-report.R#L67 unless you add `fsep = "/"` since you that path is not a local file path. You don't want to accidentally add a backslash if running on Windows!

Issue #222, opened 2023-06-15
Future thought: CNV/tumor inference


jashapiro:
If we can't do a great job with cell types, what can we do with assigning tumor/normal status?

Especially inferring CNV status.

allyhawkins:
Since we have decided to include tumor/ normal designations on the portal, I'm updating this issue to reflect the next steps. To label cells as tumor cells, we need to identify a method we like for CNV inference.  We will probably want to use [CopyKAT](https://github.com/navinlabcode/copykat/) for this, mainly because the other methods I looked at require a reference of tumor/normal, which we will not have. I think a good place to start is to use `CopyKAT` with some of the RMS libraries and compare those annotations to the submitter-provided annotations.

sjspielman:
Noting from the [RMS paper](https://doi.org/10.1016/j.devcel.2022.04.003), which used `inferCNV`

> Malignant cells were confirmed in patient tumor data by inference of copy-number variation using inferCNV version 1.1.3 of the TrinityCTAT Project (https://github.com/broadinstitute/infercnv). An average read count per gene cutoff of 0.1 was used. Reference cells were defined by those cell clusters in the Seurat analysis that expressed markers of hematopoietic or endothelial cells. Analysis included denoising, implementation of inferCNV’s i6 HMM model, and subclustering using quantile normalization.

sjspielman:
Just sharing a _highly initial result_ for one of the RMS datasets..

```
> table(sce$celltype, sce$copykat_prediction)
                       aneuploid diploid not.defined
  Fibroblast                  30      32           0
  Lymphocyte                  29      88          10
  Monocyte                    93      96           6
  Neuron                       8      81           0
  Tumor_Mesoderm              83     137           8
  Tumor_Myoblast-A            10      65           0
  Tumor_Myoblast-B             6      32           4
  Tumor_Myoblast-C           213    3196          66
  Tumor_Myoblast-D             8     109           8
  Tumor_Myocyte-A             57     483           5
  Tumor_Myocyte-B             49     583           6
  Vascular Endothelium       179      87           3
```

And at a more birds-eye view, where `submitter` is condensed to just tumor/normal for "known" RMS cell types. 

```
  submitter copykat         n
  <chr>     <chr>       <int>
1 normal    aneuploid     339
2 normal    diploid       384
3 normal    not.defined    19
4 tumor     aneuploid     426
5 tumor     diploid      4605
6 tumor     not.defined    97
[...]

PR #230, opened 2023-08-16
Explore copyKat performance on RMS data

sjspielman:
Towards #222 (but may not close it!)

For this PR, I have started a notebook to explore how `copyKat` performs on a given RMS library by comparing `copyKat` predictions (diploid aka normal vs aneuploid aka tumor) to annotations provided by project submitters. I also ignored the result files that came out of this as they are quite large (even compressed they were much too big!).

Here is the rendered notebook:
[copykat-rms-exploration.nb.html.zip](https://github.com/AlexsLemonade/sc-data-integration/files/12362906/copykat-rms-exploration.nb.html.zip)

Some things to be aware of:
- `copyKat` takes a little over an hour per run, even with all the threads I gave it. 
- The notebook I have included here was rendered after files had already been written out (you'll see an `if` to either read in a result SCE or run `copyKat` to produce it/write it out). The reason for this is the sheer amount of "unsuppressable" printed test with multiple lines of printed output per cell. Trust me, it's better this way.
- I had some trouble with `renv` here - when I went to install `copyKat`, a lot of dependencies versions weren't installing/updating very well. I have committed an updated `renv.lock` file here, but while this lockfile worked fine for me _locally_ I had trouble with it on the server! We may want to more generally bump bioc packages in this repo? Anyways, best of luck with `renv` and please share any insights with me during review :) 

I envision a good amount of review here might be discussion about next steps, too! I'm requesting a couple reviewers since I think a good amount of discussion would be, well, good!


celltype_annotation/analysis/copykat-rms-exploration.Rmd
```diff
@@ -0,0 +1,224 @@
...
+```{r}
+suppressPackageStartupMessages({
+  library(SingleCellExperiment)
+  library(copykat) # https://github.com/navinlabcode/copykat
+})
+
+# define output path and SCE file
+output_dir <-  here::here("celltype_annotation", 
+                         "analysis", 
+                         "copykat-results")
```
allyhawkins, line 32:
```suggestion
output_dir <-  here::here("celltype_annotation", 
                          "analysis", 
                          "copykat-results")
```
I think this spacing is off?

celltype_annotation/analysis/copykat-rms-exploration.Rmd
```diff
@@ -0,0 +1,224 @@
...
+  sce <- readr::read_rds(copykat_sce)
+  
+} else {
+  
+  # Otherwise, read in initial SCE and run copyKat
+  sce <- readr::read_rds(local_submitter_sce_path)
+    
+  purrr::walk(
+    c("spearman", "pearson", "euclidean"), 
+    \(x) {
```
allyhawkins, line 89:
```suggestion
    \(distance_parameter) {
```
This is silly, but with the new `purrr` we can be explicit about what we're looping over instead of using `x`. I feel like it's more helpful to have meaningful names for understanding what's happening.

celltype_annotation/analysis/copykat-rms-exploration.Rmd
```diff
@@ -0,0 +1,224 @@
...
+      
+      copykat_result <- copykat(
+        rawmat = as.matrix(counts(sce)), 
+        id.type = "E", # we have Ensembl gene ids, not the default gene symbols ("S")
+        sam.name = params$library_id, 
+        distance = x,
```
allyhawkins, line 95:
```suggestion
        distance = distance_parameter, 
```

celltype_annotation/analysis/copykat-rms-exploration.Rmd
```diff
@@ -0,0 +1,224 @@
...
+        rawmat = as.matrix(counts(sce)), 
+        id.type = "E", # we have Ensembl gene ids, not the default gene symbols ("S")
+        sam.name = params$library_id, 
+        distance = x, 
+        plot.genes = FALSE,
+        n.cores = params$threads
+      )
+      
+      # save prediction to sce
+      colname <- glue::glue("copykat_{x}")
```
allyhawkins, line 101:
```suggestion
      colname <- glue::glue("copykat_{distance_parameter}")
```

celltype_annotation/analysis/copykat-rms-exploration.Rmd
```diff
@@ -0,0 +1,224 @@
...
+      colData(sce)[[colname]] <- copykat_result$prediction$copykat.pred
```
allyhawkins, line 102: What else is output by copykat? Anything that might be worth saving besides the predictions? The contents of the [`test_output` folder on their repo](https://github.com/navinlabcode/copykat/tree/master/test_output) has a `_CNA_results.txt` matrix? Maybe that's helpful in figuring out confidence or scores?

celltype_annotation/analysis/copykat-rms-exploration.Rmd
```diff
@@ -0,0 +1,224 @@
...
+Next, we'll look at some classification metrics overall, using a custom function to calculate a confusion matrix.
+For this, we will _only consider_ cells which i) the submitter was able to classify, and ii) `copyKat` was able to classify.
+In other words, all `NA` values are removed from the data before calculations.
+
+
+
+```{r}
+# Custom function to calculate some confusion matrix quantities, including
+# accuracy, TPR, and FPR
+calculate_confusion <- function(df, test_column) {
```
allyhawkins, line 143: Is the `test_column` supposed to be the predictions column?

celltype_annotation/analysis/copykat-rms-exploration.Rmd
```diff
@@ -0,0 +1,224 @@
...
+
+# run confusion matrix function
+cat("============== SPEARMAN distance =====================")
+calculate_confusion(copykat_df, copykat_spearman)
+
+cat("============== PEARSON distance =====================")
+calculate_confusion(copykat_df, copykat_pearson)
+
+cat("============== EUCLIDIAN distance =====================")
+calculate_confusion(copykat_df, copykat_euclidian)
```
allyhawkins, line 191: yikes it looks like euclidian is definitely not the way to go!

celltype_annotation/analysis/copykat-rms-exploration.Rmd
```diff
@@ -0,0 +1,224 @@
...
+    c("spearman", "pearson", "euclidean"), 
+    \(x) {
+      
+      copykat_result <- copykat(
+        rawmat = as.matrix(counts(sce)), 
+        id.type = "E", # we have Ensembl gene ids, not the default gene symbols ("S")
+        sam.name = params$library_id, 
+        distance = x, 
+        plot.genes = FALSE,
+        n.cores = params$threads
```
allyhawkins, line 97: Are there other parameters we might want to change? Like `KS.cut`? From their docs: 
> Increasing KS.cut decreases sensitivity, i.e. less segments/breakpoints. Usually it works in a range of 0.05-0.15.

Maybe we should be playing around with that? 
-

celltype_annotation/analysis/copykat-rms-exploration.Rmd
```diff
@@ -0,0 +1,224 @@
...
+  dplyr::select(celltype, contains("copykat")) |>
+  dplyr::mutate(celltype_class = ifelse(
+    stringr::str_starts(celltype, "Tumor"), 
+    "Tumor", 
+    "Normal"
+  )) 
+
+# run confusion matrix function
+cat("============== SPEARMAN distance =====================")
+calculate_confusion(copykat_df, copykat_spearman)
```
allyhawkins, line 185: I think this is super helpful, but is there a way we could see the breakdown of accuracy per tumor/normal? I guess I'm trying to get an understanding of whether or not more normal cells are called as tumor here? Or vice versa?

celltype_annotation/analysis/copykat-rms-exploration.Rmd
```diff
@@ -0,0 +1,224 @@
...
+cat("============== PEARSON distance =====================")
+calculate_confusion(copykat_df, copykat_pearson)
+
+cat("============== EUCLIDIAN distance =====================")
+calculate_confusion(copykat_df, copykat_euclidian)
+```
+
+## Conclusions
+
+`copyKat` does not seem to agree with the submitter annotations, but possibly has the best chance of success with one of the correlational distances (spearman or pearson).
```
allyhawkins, line 196: I don't think we expect perfect agreement. I actually am pleasantly suprised by the ~ 50% accuracy.

allyhawkins (commented):
Thanks so much for getting this started! I left a few comments throughout. 
I definitely think we should try and get a measure of some sort of confidence in the calls as discussed. I don't necessarily think that needs to go in this PR though, up to you. 

I also will say that I am not totally deterred here. I really wouldn't expect perfect agreement so I feel good about this. I think if we combined this with detection of marker genes for the tumor cells then we would probably get closer to the correct results. The original paper did use a combination, so I don't expect to be totally accurate. 

Also, I haven't yet tested `renv`, but I will now and get back to you.

celltype_annotation/analysis/copykat-rms-exploration.Rmd
sjspielman: All of that output is contained within the `copyKat` object itself that I saved (but did not commit!), including that matrix. It doesn't appear to be terribly useful, though. From the README - 

> The first 3 columns in the CNV matrix are the genomic coordinates. Rows are 220KB bins in genomic orders. It also output CNVs indexed by gene names and other information such as chromosome names, start and end positions, G staining band etc.

In addition to those first 3 columns, it then has 1 column _per cell_ with information for each 220KB bin. Here's a bit of what it looks like, just showing 2 cells. It appears to mostly (only?) be useful for making the heatmap, which doesn't relate to our needs. 

```
> result$CNAmat[1:20, 1:5]
   chrom chrompos  abspos GACTCTCCACAAGCTT TTGCCTGGTTTACTTC
1      1  1042457 1042457      -0.06569387       0.02518337
2      1  1265484 1265484      -0.06569387       0.02518337
3      1  1519859 1519859      -0.06569387       0.02518337
4      1  1826619 1826619      -0.06569387       0.02518337
5      1  2058465 2058465      -0.06569387       0.02518337
6      1  2280372 2280372      -0.06569387       0.02518337
7      1  2491263 2491263      -0.06569387       0.02518337
8      1  2817807 2817807      -0.06569387       0.02518337
9      1  3025418 3025418      -0.06569387       0.02518337
10     1  3230151 3230151      -0.06569387       0.02518337
11     1  3437000 3437000       0.05664600      -0.10335066
12     1  3645603 3645603       0.05664600      -0.10335066
13     1  4091348 4091348       0.05664600      -0.10335066
14     1  4312450 4312450       0.05664600      -0.10335066
15     1  4522391 4522391       0.05664600      -0.10335066
16     1  4733807 4733807       0.05664600      -0.10335066
17     1  4949751 4949751       0.05664600      -0.10335066
18     1  5167781 5167781       0.05664600      -0.10335066
19     1  5378577 5378577       0.05664600      -0.10335066
20     1  5586352 5586352       0.05664600      -0.10335066
```

celltype_annotation/analysis/copykat-rms-exploration.Rmd
sjspielman: yes, I can rename!

celltype_annotation/analysis/copykat-rms-exploration.Rmd
sjspielman: Ha I don't know if you meant o emphasize this text but I love that that's how it happened! Let's see what this param can do for us using only one of the correlational distances.

allyhawkins: Haha oh boy I did not mean to do that!

celltype_annotation/analysis/copykat-rms-exploration.Rmd
sjspielman: Sure it certainly could be worse, but also 50% accuracy = pretty much random chance 😬

jashapiro: Spearman and Pearson have >80% FPR though, so I don't think they doing well either.

sjspielman:
Sharing one quick code-spelunking insight - the "WARNING! NOT CONVERGENT!" business appears _not to be related_ to the actual classification of cells as diploid vs aneuploid, but it emerges from the copyKat step that attempts to define a neutral/diploid baseline that will scaffold subsequent calculations. So, capturing this information won't be informative as a final confidence score.

sjspielman:
Alright folks, I'm back with a ~substantial update to the notebook! 

Notebook can be viewed here, and currently contains...
[copykat-rms-exploration.nb.html.zip](https://github.com/AlexsLemonade/sc-data-integration/files/12383478/copykat-rms-exploration.nb.html.zip)

- To start, 
  - A brief overview of the `copyKat` algorithm (largely based on how it is described in the paper)
  - A couple function definitions which are used to assess results
- Part 1: compare distance parameters (punchline: do not use default "euclidean" distance)
- Part 2: compare KS.cut parameters (punchline: default value seems fine)
- Conclusions section has been updated to match these results as well, with a couple more thoughts

The notebook also exports the main results as TSV files which are now under version control, and will make it easier for others to run the notebook without having to wait 10 hours..! With these results "cached," it knits up ASAP. But, to further help with review and further development as needed, I also added a "test" parameter - when set to `TRUE`, the SCE will be subsetted to only 50 cells which allows for a much faster runtime; with 2 cores, each inference takes about 4-5 minutes, so the whole notebook can be run _from scratch_ in about an hour. 

One bonus of implementing that test condition was that I discovered something important about `copyKat` output related to our discussions about getting confidence measures. I learned that when you give `copyKat` insufficient data (aka, a test dataset with 50 cells), it will indeed report that predictions are "low confidence" [(source)](https://github.com/navinlabcode/copykat/blob/b795ff793522499f814f6ae282aad1aab790902f/R/copykat.R#L413-L421). That seems to specifically occur because the normal diploid baseline was not confidently inferred, hence predictions relying on that baseline cannot be confidently inferred. So, since the full SCE did not give these "low.conf" annotations, it implies that `copyKat` felt confident about the predictions!

I'll also note that I think this notebook is possibly already a little too big. I thought about splitting it up into 2, but decided 1 notebook to explore parameters (as long as that's _all_ it does!) is possibly ok! But, we should move into a 2nd notebook for any next steps, I think.

jaclyn-taroni:
From a cursory glance, it seems the additional metrics tell us that if we used CNV inference from copyKAT _only_, it'd be worse than a random classifier. So, won't be doing that!

celltype_annotation/analysis/copykat-rms-exploration.Rmd
```diff
@@ -0,0 +1,433 @@
...
+
+## Background: `copyKat` method
+
+Reference: https://doi.org/10.1038/s41587-020-00795-2
+
+To begin, here's a brief overview of the `copyKat` approach, parly quoting the paper throughout.
+See Figure 1 in the paper for a method diagram:
+
+* `copyKat` takes the _raw expression matrix_ and stabilizes variance, smooths the outliers
+* A subset of cells is then identified to serve as a diploid baseline for subsequent identification of diploid/aneuploid
```
allyhawkins, line 29: Is this information returned at all? Can you go back and see which cells are being classified as the "confident diploid"?

celltype_annotation/analysis/copykat-rms-exploration.Rmd
```diff
@@ -0,0 +1,433 @@
...
+  compare_result_tables("euclidean distance") 
+
+copykat_df |>
+  dplyr::select(submitter, submitter_binary, copykat = copykat_pearson) |>
+  compare_result_tables("pearson distance") 
+
+copykat_df |>
+  dplyr::select(submitter, submitter_binary, copykat = copykat_spearman) |>
+  compare_result_tables("spearman distance") 
+
```
allyhawkins, line 292: 1. Can you create a single vector of the distances and `purrr::map` across that to simplify this? 
2. I think I would add some concluding statements here on how you are interpreting these results. Like a statement about the negative kappas throughout and any other metrics that stand out.

celltype_annotation/analysis/copykat-rms-exploration.Rmd
```diff
@@ -0,0 +1,433 @@
...
+For this, we'll start with the SCE from the last section, and build it up from there with other `KS.cut` values.
+Since it was already run with default `KS.cut = 0.1`, we'll test a few additional values here and compare those to above.
+
+```{r}
+if (file.exists(copykat_kscut_tsv)) {
+  # if the TSV exists, read it in
+  copykat_df <- readr::read_tsv(copykat_kscut_tsv)
+
+} else if (file.exists(copykat_distance_tsv)) {
```
allyhawkins, line 315: I do think you could create a function that takes as input the tsv file and any parameters at the beginning of the notebook and then you won't have to rewrite this function again. Basically remove the `purrr` statement from here, make any parameters you want to test arguments with default values and then use `purrr` to run the function for the range of values for the parameter you are interested in. But I also don't know how much time we want to continue to spend on this.

allyhawkins (commented):
Thanks for making these updates. Generally this doesn't look good. For me, the bar charts are helpful as a visual, so thanks for adding that! A few overall comments: 

- Can you add some more concluding comments after each of the print outs of tables/ confusion matrix results? I think it's helpful to have an explanation of what you're concluding, what the metrics are telling you, etc. 
- If you wanted to clean things up then it would probably benefit from creating a function for running kappa. I left a comment about it, but you have almost the exact same code twice just testing different parameters. 
- I think if we were going to keep exploring this then it would be helpful to pull out some of the key metrics from running the confusion matrix and plot those for easier comparison across the distance metrics and ks.cut values. However, I don't know if it's worth it at this point because I'm not feeling great about using this. 
- I am curious which cells here are defined as the "confident diploid" baseline that's used. Are they all normal cells? Because my guess is no and I'm not sure our datasets will fit these assumptions.

sjspielman:
> I am curious which cells here are defined as the "confident diploid" baseline that's used. Are they all normal cells? Because my guess is no and I'm not sure our datasets will fit these assumptions.

This information does not appear to be exported. From their code, this is where the baseline is assessed, under the condition where no known normal cells are provided:
https://github.com/navinlabcode/copykat/blob/b795ff793522499f814f6ae282aad1aab790902f/R/copykat.R#L163-L181

I think we'd need to fork the repo and make changes ourselves to export this information, but that may not be worth our time.

celltype_annotation/analysis/copykat-rms-exploration.Rmd
sjspielman: I wish 🙁 
https://github.com/AlexsLemonade/sc-data-integration/pull/230#issuecomment-1686712505

celltype_annotation/analysis/copykat-rms-exploration.Rmd
sjspielman: I thought about doing this actually when writing the notebook, but I didn't because it won't make much of a difference in terms of repeated code b/c each time we run `copyKat`, we vary a _different_ set of params. So, this still needs to have an `if` in this function to determine how to run (e.g. ,should it be `distance = the purrr variable to loop over` or `ks.cut = purrr variable to loop over`). 

That said, I agree it will clean up the notebook overall and make it easier to hone into relevant parts, so I'll go for it now!

sjspielman:
> I think if we were going to keep exploring this then it would be helpful to pull out some of the key metrics from running the confusion matrix and plot those for easier comparison across the distance metrics and ks.cut values. However, I don't know if it's worth it at this point because I'm not feeling great about using this.

How about I output this information so that we can continue in a second notebook if we decide to?

allyhawkins:
> > I think if we were going to keep exploring this then it would be helpful to pull out some of the key metrics from running the confusion matrix and plot those for easier comparison across the distance metrics and ks.cut values. However, I don't know if it's worth it at this point because I'm not feeling great about using this.
> 
> How about I output this information so that we can continue in a second notebook if we decide to?

Yes I like that idea!

celltype_annotation/analysis/copykat-rms-exploration.Rmd
sjspielman: Noting I've simplified this to use `purrr`, but concluding statements are forthcoming!

sjspielman:
@allyhawkins I believe I've hit all your comments here! Some of the code can still probably be simplified in a few places, but I think this is a reasonable level for how we intend to use this notebook going forward.

Here's the current HTML!
[copykat-rms-exploration.nb.html.zip](https://github.com/AlexsLemonade/sc-data-integration/files/12410271/copykat-rms-exploration.nb.html.zip)

allyhawkins (approved):
I think this looks good to me at this stage. Thanks for adding in the conclusions and interpretation comments. I would move that section, but other than that this is good to go.

sjspielman:
Thanks for all the feedback here @allyhawkins! Will make that change and move this along :)

